PR #7, opened 2019-07-24
Organize as python package, fixes

JelleAalbers:
This will
  * Make this repository an installable python package. I assumed 'strax_interface' was the main code, and put everything else in a 'stuff' folder for now. I hope this makes sense: I'm not sure how you were organizing the development, but it looked like Peter and Tianyu were working in separate folders. I did not add any unit tests in the `tests` folder, but the infrastructure to do so is now in place.
  * Add a notebook showing how to call the simulators from a strax context.
  * Let the strax plugins override `compute` (as other plugins) rather than `iter`. The latter is a low-level strax method that does other things you were now missing out on. Also fixed a small bug where run_id was not defined.
  * Make PeaksFromFax and RawRecordsFromFax inherit from a common class so their long option-definition and parsing does not need to be redefined..
  * Replace the large files you had to copy separately with small files that are auto-downloaded from our aux file repo, see below.
  * tqdm(enumerate(...)) -> enumerate(tqdm()) so the progressbars work.

Unless I'm misunderstood the code (which is quite possible) there were some issues with the PeakSimulator code:
  * It tried to iterate over a single instruction, i.e. loop over the fields event_number, type, t, x, etc. and simulate a single peak for each. This is completely mad and also crashes a few lines into the loop. I fixed it by letting it make one peak per instruction, but I think this might still gobble up the peak and its afterpulses into one peak.
  * I'm not sure it actually bypasses any of fax's raw waveform simulation. It looks more like it simulates the waveform as usual, then makes a strax peak immediately from the simulated pulses. That indeed skips over strax's low-level pulse processing, but the slow part we would actually like to skip is the photon-by-photon, PMT-by-PMT generation of the pulses.

There are also some strange hangs occuring. In the RawRecordsSimulator this seems to happen when there is a chunk without pulses, in PeakSimulator I'm not sure.

Reducing the data file sizes
-----------------------------------

**Noise file**

Keep 10% of the noise data and use savez_compressed instead of savez to save it; 200 -> 2 MB.

```python
noise_bla = straxen.get_resource('170203_0850_00.npz', fmt='npy')['arr_0']
np.savez_compressed('x1t_noise_170203_0850_00_small.npz', noise_bla[:10000, :])
```

**Electron afterpulse file**

Actually this does reduce the filesize (from 9 -> 0.5 MB), but the main point of this change is to simplify the code.

The electron afterpulse configuration contained ``np.array({'liquid': {'delaytime_cdf': x}, dtype='object')``, with `x` an array of 800 000 numbers describing the delay time CDF with ns precision. Single electrons are ~200 ns objects; we don't need to localize them with 1 ns precision. I replaced them with a 10-ns resolution probability mass function (diff of the CDF):

```python
import gzip
import pickle

from multihist import Hist1d

x = np.load('ele_after_pulse.npy').item()
x = x['liquid']['delaytime_cdf']
h = Hist1d.from_histogram(
    histogram=np.diff(np.concatenate([x, [x[-1]]])).reshape(-1, 10).mean(axis=1),
    bin_edges=np.arange(0, len(x) + 1, 10))
h.plot()
with gzip.open('x1t_se_afterpulse_delaytime.pkl', mode='wb') as f:
    pickle.dump(h, f)
```
I use multihist to store it, so we can just call .get_random() on it to sample delay times.

**PMT afterpulse file**

You might be able to do something similar here, but I did not change the code. Converted all dtypes to float32, then saved as gzip-compressed pickle. 220 -> 22 MB. 

```python
data = np.load('pmt_after_pulse.npy').item()

# Convert to float32
for aptype, d in data.items():
    for key, arr in d.items():
        data[aptype][key] = arr.astype(np.float32)
        
# Save as compressed pickle
with gzip.open('x1t_pmt_afterpulse_config.pkl.gz', mode='wb') as f:
    pickle.dump(data, f)
```


zhut19:
Thanks a lot @JelleAalbers ,
 - The code in strax interface are the main code, what's now in stuff are some optimizing and clarification of subparts that need to pr into the main code.
 - There are some issues in the fax.py needs to be fixed as well.
 - Not sure bypassing photon-by-photon generation is the end goal here, at least we should try our best to speed those up first.

petergaemers:
Ha @JelleAalbers, this looks awesome.

As for the Peak Simulator
* It is possible I changed some things from RawRecords causing the Peak simulator to crash regarding the instructions.
*It does not bypass the PMT-by-PMT pulse generation. Simulating in different PMTs or everything in 1 PMT did not make a significant difference in computation time. The computation time is currently dominated by the double photo-electron emission and the Pulse.add_current function. According to the new numba version compilation of this function fails so this should be improvable

PR #9, opened 2019-07-28
Add tests, njit add_current

JelleAalbers:
This will:
  * Add a unit test that verifies the simulator does not crash.
  * Add a check in the strax interface that chunks of raw_records passed to straxen are sorted by time and sufficiently well-spaced. 
    * Strax, or rather the low-level processing algorithms in straxen, require(s) this. The DAQReader does work for to ensure it, but here we're not reading from the DAQ.
    * If you pass unsorted or (nearly) overlapping chunks of records to strax it can cause a hang until the mailbox timeouts expire (I'm not exactly sure why). 
    * I did not want to dig too deeply to figure out why we got non-sorted or overlapping chunks in some conditions, so I just added a check and made it throw an exception if we see the problem. I hope a future fix would solve the problem and remove the check.
  * Replace the `guvectorize`d `add_current` with a `njit`ed explicit for loop. This gives about a 30% speedup to the simulator (tested by simulating 1000-electron S2s). 
    * The problem was that numba couldn't figure out the types of the globals that the guvectorized function used, and thus used its slower object mode rather than the main nopython mode.
    * I didn't actually intend to do any optimization, but a nice warning message about this issue from numba 0.45 pointed out the problem very specifically. Apparently numba is retiring object mode; in the future numba will just give an error in cases like this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 dist: xenial
+cache: pip
 python:
 - '3.6'
 - '3.7'
--- a/tests/test_strax_interface.py
+++ b/tests/test_strax_interface.py
@@ -1,0 +1,21 @@
+import strax
+import straxen
+import wfsim
+
+strax.mailbox.Mailbox.DEFAULT_TIMEOUT = 20
+
+
+def test_sim():
+    # Just some id from post-SR1, so the corrections work
+    run_id = '180519_1902'
+
+    st = strax.Context(
+        register=wfsim.RawRecordsFromFax,
+        config=dict(nevents=4),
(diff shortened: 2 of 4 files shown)


JelleAalbers:
Just saw @petergaemers already flagged the guvectorize problem in https://github.com/XENONnT/WFSim/pull/7#issuecomment-515015117

petergaemers:
Hi @JelleAalbers 
Thank you very much! (You beat me to it :/)

* As for the overlapping pulses. All pulses from different sources are treated separately. So things like a PMT afterpulse of the S1 overlapping with the S2 are still possible

PR #26, opened 2019-10-18
Output to pax readable files

zhut19:
Renewed pax_interface.py so the current framework can run and save data into something pax can use (with some additional steps when calling pax).


zhut19:
On the pax side see [code](https://gist.github.com/zhut19/b180ea2aca9fbf210da5dfbfc02d5524)

PR #27, opened 2019-10-23
Update G4 and PAX interface

HenningSE:
New function to read Geant4 files. We are now using the individual positions of each interactions and the corresponding energy depositions. The event number is taken from the Geant4 root file as well. The events are now separated by a constant time difference. I think this can be changed in the future to be more efficient. 

Things to do next:
-  get the recoil type (ER, NR) from the interaction
-  get the interacting particle for NEST from the Geant4 file
- do we need some clustering like in nSort?
- Why are there cy > 1 values in core.py?

### DIFF
--- a/wfsim/__init__.py
+++ b/wfsim/__init__.py
@@ -2,3 +2,4 @@ __version__ = '0.0.3'
 
 from .core import *
 from .strax_interface import *
+from .pax_interface import *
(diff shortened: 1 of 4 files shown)


petergaemers:
Hi Henning,

Thank you very much.
Regarding nSort like clustering, I do not know if it really makes sense to copy half of nSort into here.

PR #28, opened 2019-11-11
Update strax_interface.py

petergaemers:
This will fix:
 * The horrible chunking bug where we can only write data to 1 chunk before everything crashes


zhut19:
Good catch, thanks

PR #29, opened 2019-11-13
Redesign chunk mechanism

zhut19:
Event_number in instruction was used as a temporary solution to chunking, however, event_number  is intended to be reserved for other purposes. While chunk should be done automatically by 'time'.

This pr also try to solve 
1) Crash due to unable to yield when event_rate is set above 100 Hz and cut off time set > 0.1 seconds. 
2) Long term effect e.g. photo ionization (PI) electrons are saved into instruction buffer as well as the truth.
3) Double photon electron is propagated to the truth output as well.

Immediate effects are:
1) Instruction input can be unsorted now
2) Must specify chunk_size in config, (especially when not using build-in random instruction)
3) Running might be just a tiny bit slower.

For the future:
1) Delay coincidence can be implemented very easily.
2) Long term single electron effect can be added the same way as PI.

### DIFF
--- a/wfsim/core.py
+++ b/wfsim/core.py
@@ -72,6 +72,10 @@ class Pulse(object):
                 # Add some double photoelectron emission by adding another sampled gain
                 n_double_pe = np.random.binomial(len(_channel_photon_timings),
                                                  p=self.config['p_double_pe_emision'])
+                self._n_double_pe += n_double_pe
+                if channel in self.config['channels_bottom']:
+                    self._n_double_pe_bot += n_double_pe
+
                 _dpe_index = np.random.choice(np.arange(len(_channel_photon_timings)),
                                               size=n_double_pe, replace=False)
                 _channel_photon_gains[_dpe_index] += self.config['gains'][channel] \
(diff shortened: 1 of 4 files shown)


zhut19:
Something strange happened during testing this branch.
s2_linear
s1_linear
The reconstructed peak area seems to be twice as big as the truth peak area. Not sure this is a problem in wfsim, maybe more likely to be something in straxen?

HenningSE:
Hi tianyu, i was trying to run this changes in combination with the pax interface but it seems to be, that the `rawdata.instruction_index` is missing now, which is still used in the `PaxEvents` class.

zhut19:
Thanks, @HenningSE for checking! The pax interface should be working now.

HenningSE:
Thank you for the implementation @zhut19, it is working fine!

HenningSE:
> Something strange happened during testing this branch.
> s2_linear
> s1_linear
> The reconstructed peak area seems to be twice as big as the truth peak area. Not sure this is a problem in wfsim, maybe more likely to be something in straxen?

I can reproduce this problem with Kr83m simulations and using PAX for processing. 
S1a_for_pr
S2a_for_pr

wfsim/core.py
```diff
@@ -612,91 +607,147 @@ def __init__(self, config):
...
-            self.pulses['pmt_ap'](self.pulses['ele_ap'])
-            self.pulses_cache += getattr(self.pulses['pmt_ap'], '_pulses')
-
-        if len(self.pulses_cache) > 0:
-            self.last_pulse_end_time = max(self.last_pulse_end_time,
-                np.max([p['right'] for p in self.pulses_cache]) * 10)
-
+        
+        for pt in [ptype, 'pmt_ap']:
+            _pulses = getattr(self.pulses[ptype], '_pulses')
```
zhut19: Should be 'pt' as the key, the double-counting is coming from here.

HenningSE: Good catch! I did some fast simulations and it looks promising! 
S1a_for_pr
S2a_for_pr
Only the S1_b seems to be still off.... (was the same before the redesign of the chunk mechanism)
S1b_for_pr

PR #31, opened 2019-11-21
Add features required for XENONnT configuration

petergaemers:
This pull requests add the needed things to run the waveform simulator with a XENONnT configuration.
* Open issue: light yield maps crash during interpolation so now are fixed values
* Some files are to large to host on github, so we need to put them somewhere else and the paths in strax_interface and load_resource need to be set by hand

*Strax also doesn't work with a nT configuration at the moment, so some changes to pulse_processing in strax are also needed before it works

### DIFF
--- a/wfsim/core.py
+++ b/wfsim/core.py
@@ -1,7 +1,6 @@
 import logging
-import time
 
-from numba import int64, float64, njit
+from numba import njit
 import numpy as np
 from scipy.interpolate import interp1d
 from tqdm import tqdm
@@ -229,7 +228,11 @@ class S1(Pulse):
             np.array(v).reshape(-1) for v in zip(*instruction)]
         
         positions = np.array([x, y, z]).T  # For map interpolation
-        ly = self.resource.s1_light_yield_map(positions) * self.config['s1_detection_efficiency']
+        if self.config['detector']=='XENONnT':
+            #for some reason light yield map crashes TODO
+            ly = 1
+        else:
+            self.resource.s1_light_yield_map(positions) * self.config['s1_detection_efficiency']
         n_photons = np.random.binomial(n=n_photons, p=ly)
 
         self._photon_timings = np.array([])
@@ -319,7 +322,11 @@ class S2(Pulse):
             np.array(v).reshape(-1) for v in zip(*instruction)]
 
         positions = np.array([x, y]).T  # For map interpolation
-        sc_gain = self.resource.s2_light_yield_map(positions) * self.config['s2_secondary_sc_gain']
+        if self.config['detector'] == 'XENONnT':
+            #Light yield map crashes, but the result is 1 for all positions in the tpc
+            sc_gain = np.repeat(self.config['s2_secondary_sc_gain'], len(positions))
+        else:
+            sc_gain = self.resource.s2_light_yield_map(positions) * self.config['s2_secondary_sc_gain']
 
         # Average drift time of the electrons
         self.drift_time_mean = - z / \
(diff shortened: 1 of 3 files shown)


wfsim/core.py
```diff
@@ -809,6 +831,8 @@ def get_truth(self, instruction, truth_buffer):
                 truth_buffer[ix]['t_first_{name}'.format(name=name)] = np.nan
                 truth_buffer[ix]['t_last_{name}'.format(name=name)] = np.nan
                 truth_buffer[ix]['t_sigma_{name}'.format(name=name)] = np.nan
+            truth_buffer[ix]['endtime'] = np.max(
```
zhut19: t_last_photon should be the endtime already, not sure why the electrons need to be added in here.

wfsim/load_resource.py
```diff
@@ -47,24 +43,36 @@ def __init__(self, config={}):
...
+                # S2
+                self.s2_light_yield_map = itp_map(self.config['s2_light_yield_map'], fmt='json')
                 self.s2_per_pmt_params = get_resource(self.config['s2_per_pmt_params'], fmt='csv')
-            if config['experiment']=='XENONnT':
-                self.s2_per_pmt_params = itp_map(self.config['s2_pattern_map'], fmt='json.gz')
-            # Electron After Pulses
-            # self.uniform_to_ele_ap = get_resource(self.config['ele_ap_cdfs'], fmt='npy')
+
+            if config['detector'] == 'XENONnT':
+                self.s1_light_yield =  get_resource('/Users/petergaemers/Downloads/XENONnT_s1_xyz_patterns_corrected_MCv3.0.0.json.gz',fmt = 'json.gz')
```
zhut19: It's strange that lce and pattern are using the same map?

zhut19:
Hi, Peter. Thanks for gathering all the maps for nT. Can the file size issue be solved by compress them more? Otherwise, probably they can also be loaded by everyone if you put a copy of them directly in those public folders. Unfortunately, the load resource function used here is a bit different from the one in straxen, so the file name is the real name instead of a deterministic hash [see here](https://github.com/XENONnT/WFSim/blob/1f5b5e9f5e1598c1f386be9343232358f0cb1136/wfsim/utils.py#L49). The  reason I didn't want to use the deterministic hash is because it's built too deep in straxen and wfsim should be mostly independent of straxen.

wfsim/core.py
```diff
@@ -704,6 +722,10 @@ def symtype(ptype):
         return ['s1', 's2', 'unknown', 'pi_el', 'pmt_ap'][ptype - 1]
 
     def sim_data(self, instruction):
+        if len(instruction) > 8:
```
zhut19: Ah, I forgot about this part. Thanks for adding it back!

wfsim/strax_interface.py
```diff
@@ -311,5 +309,5 @@ def compute(self, chunk_i):
             result = next(self.sim_iter)
         except StopIteration:
             raise RuntimeError("Bug in chunk count computation")
-        # self._sort_check(result['raw_records'])
+        self._sort_check(result['raw_records'])
```
zhut19: Do you know why there should be a gap in between chunks? Currently, the result from ChunkRawRecords are already guaranteed to be sorted in time.

wfsim/core.py
petergaemers: Endtime is wanted by strax: https://strax.readthedocs.io/en/latest/advanced/plugin_dev.html
Adding this makes it possible to include the truth in a loopplugin making matching really easy

wfsim/load_resource.py
petergaemers: I asked Lutz. The LCE map is the sum over all channels in the pattern map. And it's a bit redundunt to have two maps with the same information

wfsim/strax_interface.py
petergaemers: Strax expects a minimum time distance between chunks. For details why we need @JelleAalbers.
Currently the chunks are sorted, however the signals on the border gets written to two seperate chunks. This is not what we want

wfsim/load_resource.py
l-althueser: The one addition that you get from the light yield map is the pe/kev value. If you make the light yield relative it is identical to the relative LCE (but from MC of course, instead of data). So if you anyway using the light yield map just for relative values it's fine ... Otherwise we have to find another approach here ;)

l-althueser:
The one addition that you get from the light yield map is the pe/kev value. If you make the light yield relative it is identical to the relative LCE (but from MC of course, instead of data). So if you anyway using the light yield map just for relative values it's fine ... Otherwise we have to find another approach here ;)

l-althueser:
Just wanted to add here that there is an additional downside with using GitHub for maps. Some computing nodes/cluster don't have internet access - so we should have at least the option to also point to a local directory in the future ;)

zhut19:
Yes, I do prefer a github implementation as well. During testing, my notebook running on xenon1t nodes usually doesn't have access to github. Thus currently any of the following folders should also work. 
['./resource_cache', '/tmp/straxen_resource_cache',  '/dali/lgrandi/strax/resource_cache']
Additionally unlike straxen that files are accessed via hash string, the get_resource in the wfsim will locate the maps by (human-readable) name.

l-althueser:
Actually a way to have versioned maps without the file size limitation of GitHub would be good. Isn't this the same/similar issue that strax will experience once the production phase has completely started?

wfsim/strax_interface.py
JelleAalbers: Strax expects that it can find peaks independently in each chunk, and peaks may never overlap in strax. If there is no gap between chunks, the peaks can overlap (depending on the left and right extension set in the peakfinder / hitfinder).

petergaemers:
I've put all the files on /dali/lgrandi/pgaemers/fax_files and updated the paths.
For versioning we need to think of something clever though

PR #36, opened 2019-12-24
Straxen integration and test build fix

JelleAalbers:
This fixes the Travis build and improves integration with straxen.
  * Make WFsim use straxen's functions for loading resources and building maps. I moved several of the improvements you made (thank you!) to the corresponding functions in straxen. (If you make straxen pull requests for these in the future, we will all benefit from the improvements.)
  * Load the Resource class using a separate `load_config` function, which keeps a cache to avoid reloading the same config. The current implementation with a singleton class would only allow loading one configuration per python session -- if I'm not mistaken.
  * Add a XENONnT simulation test.
  * Fix default paths to point to github, not /dali. 
  * Fix reading instructions from csv. We now explicitly check that the csv produces the correct dtype. If we wouldn't do this, you'd get deeper errors later (cannot convert int64 to float etc).

To fit the large XENONnT per-PMT MC maps into strax_auxiliary files, I quantized them to values of 1e-5. That is, I store the maps as 16-bit integer multiples of 1e-5, instead of 64-bit floats. I also changed the compression of the maps to zstd instead of gzip, to improve the loading speed. Here is the code I used to quantize and compress the maps:

```python
import gzip
import json
import os.path as osp
import pickle

import numpy as np
import straxen
import zstd

filename = 'XENONnT_s1_xyz_patterns_corrected_MCv3.0.0'
orig_format = 'json.gz'
orig_dir = '/dali/lgrandi/pgaemers/fax_files/'

fn = osp.join(orig_dir, filename) + '.' + orig_format
print(f"Loading original file {fn}, {osp.getsize(fn)/1e6:.2f} MB")
data = straxen.get_resource(fn, fmt=orig_format)

print("Creating quantized map")
mp = np.array(data['map'])
quantum = 0.00001 
dtype = np.uint16
encode_until = np.iinfo(dtype).max * quantum
if mp.max() > encode_until:
    raise ValueError(f"Map maximum value is {mp.max():.4f}, can encode values until {encode_until:.4f}")
    
data['quantized'] = quantum
data['compressed'] = ('zstd', dtype, mp.shape)
data['map'] = zstd.compress(np.round(mp / quantum).astype(dtype))

fn = filename + '.pkl'
with open(fn, mode='wb') as f:
    pickle.dump(data, f)
print(f"Wrote new file {fn}, {osp.getsize(fn)/1e6:.2f} MB")
```

### DIFF
--- a/notebooks/Fax_Tutorial.ipynb
+++ b/notebooks/Fax_Tutorial.ipynb
@@ -19,7 +19,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -51,14 +51,23 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "url_base = 'https://raw.githubusercontent.com/XENONnT/strax_auxiliary_files/master/fax_files/'"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
     "st = strax.Context(\n",
     "    register=wfsim.RawRecordsFromFax,\n",
     "    config=dict(detector=\"XENONnT\",\n",
-    "               fax_config = '/dali/lgrandi/pgaemers/fax_files/fax_config_nt.json',\n",
+    "                fax_config=url_base + 'fax_config_nt.json'),\n",
     "    **straxen.contexts.common_opts)"
    ]
   },
@@ -72,7 +81,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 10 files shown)


JelleAalbers:
You can now override configuration options by setting `fax_config_override` to a dictionary with options in your strax context config, for example:
```python
st.get_data('peaks', config=dict(fax_config_override=dict(
    electron_extraction_yield=0)))
```
or
```python
st.set_config(dict(fax_config_override=dict(
    electron_extraction_yield=0)))
st.get_data(...)
```

Setting instruction-by-instruction config options through monkeypatching the instruction dtype, as explained in the [advanced tutorial](https://github.com/XENONnT/WFSim/blob/master/notebooks/Advanced_tricks.ipynb), is still possible.

I removed the "experiment" option from the strax interface, it seemed to be unused, and there was also a "detector" option.

zhut19:
Thanks a lot, Jelle. Those look like a lot of improvements.
  * Great that you moved the loading resource and building maps into straxen, I should do that a while ago. I remember there're some slight differences from the one from [s2_only itp_map](https://github.com/XENON1T/s2only/blob/master/s2only/itp_map.py) for including multiple pmt channels. That was done quite a while ago, I should go check it's actually there and also if you noticed it.

 * The singleton class is used for avoiding [multiple instances in core.py](https://github.com/XENONnT/WFSim/blob/f1dd00e3bba3bbd2c88bdd2104be5d506d1f6193/wfsim/core.py#L639) of Resource class where all the map interpolation functions are cached. If I am not misunderstanding this pr, you would just cache the config (where maps are really just strings).

Thanks again for the awesome updates, let me pull this and check it out before merging.

JelleAalbers:
Hi Tianyu, thanks for having a look!

Yes, the extension to cover vector/array valued maps is included in [straxen](https://github.com/XENONnT/straxen/blob/master/straxen/itp_map.py)'s map now, as is support for your more compact specification of the coordinate system.

The `load_config` does actually cache `Resource` instances, not configs, so perhaps the name `cached_configs` is misleading (I guess I viewed `Resource` as a kind of 'initialized config'). We make a hash of the config (where the maps are strings) and use it as a key to store and recall a cached `Resource` instance (where the maps are loaded). This way you will have only one `Resource` instance active if you only use one config, but you will be able to run two contexts with different simulation settings, configs, and Resource instances next to each other.

PR #37, opened 2020-01-04
Truth writing fixes

JelleAalbers:
This fixes a few things in the truth info writing.
  * Due to an indexing quirk, the truth buffer was never actually marked as cleared in `final_results` of `ChunkRawRecords`. As a result, there were many duplicate entries in the truth output for long datasets.
  * There was a [mutable default](https://docs.python-guide.org/writing/gotchas/) for `truth_info` in RawData. I don't think this is actually used, so perhaps we can just remove it instead of using the None-default trick?
  * Maybe this was a feature, not a bug, but for afterpulse-like peak the t_first_electron etc. fields were not set to NaN when there are no electrons. Since the truth buffer gets reused without being cleared, I think this would lead to values from old peaks showing up in the truth for new peaks. Just let me know if I misunderstood.
  * There was an undefined variable 'pulse' in the handling of channel-specific self-trigger / ZLE thresholds. This isn't related to truth file writing; I just saw it accidentally because it's right next to it. I hope I put the right variable in now, but I don't know -- the test apparently don't hit this code.

(I could also not resist a slight cleanup / annotation of the truth writing code while trying to understand it).

(The Travis build will fail until we apply the fix in #36)

### DIFF
--- a/notebooks/Fax_Tutorial.ipynb
+++ b/notebooks/Fax_Tutorial.ipynb
@@ -19,7 +19,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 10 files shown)


zhut19:
Regarding the afterpulse-like peak without clearing the tb. The idea is to avoid saving 0-photon secondary signals. Which can happen frequently due to electron lifetime being implemented at pulse generation class.  I guess the potential problem you thought of can happen when `quantum` loop through `photon` and hit the afterpulse return during `electron`.
  * Currently, this shouldn't be an issue, due to that I am not exporting photon-only afterpulse truth info.
(`get_truth` is only called upon elements in the instruction list, and since pmt-ap doesn't write anything to this list, pmt-ap won't get into the truth buffer)

 * But the `if` should be more clear as shouldn't break during the middle of the loop:
```
if (peak_type not in ['s1', 's2'] 
    and quantum == 'photon'): return
```

PR #38, opened 2020-01-09
edit truth handing for non-detected signals

jorana:
This fix is implemented by @petergaemers.

For signals that were not detected, no truth info was written as the condition on line https://github.com/XENONnT/WFSim/blob/02bcef02c83b42af5daae1df969913078dbba9a2/wfsim/strax_interface.py#L226 was never satisfied for np.nans.

This resulted in fewer S1's in the truth than we requested in the instructions.


zhut19:
Thanks for updating this, you are right that truth shouldn't ignore zero-photon signals.

PR #40, opened 2020-02-12
Allow config override of files

jhowl01:
Amends load_resource.py to allow different filenames for light yield maps, etc to be passed via the config. Default hard-coded files are preserved. Unity light yields are helpful for understanding reconstruction efficiencies.

Note: I also added "placeholder_map.json" to strax_auxilliary_files. I think you could try this as the default for nT instead of doing [this](https://github.com/XENONnT/WFSim/blob/cce8748785cf1a208c44be6df84655b77ad60f80/wfsim/core.py#L247) but I didn't want to mess with it.


petergaemers:
Hey Joey,

Thank you very much

PR #42, opened 2020-03-11
Catch up strax

zhut19:
The strax has updated to [v0.9.0](https://github.com/AxFoundation/strax/releases/tag/v0.9.0), requiring 
chunk class (dictionary of chunk classes) for return.

Other updates:
  - Remove gains from config, instead use to_pe file to calculate gain.
  - Add the secondary signal from the photo-electric effect of the gate, only for S2. Matching S2 pulse shape by eye.
  - Fix some issues with truth output when there're dead pmts.
  - Update pax_interface so it can work again.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ with open('HISTORY.md') as file:
 
 setuptools.setup(
     name='wfsim',
-    version='0.0.3',
+    version='0.0.4',
     description='XENONnT Waveform simulator',
     author='Wfsim contributors, the XENON collaboration',
     url='https://github.com/XENONnT/wfsim',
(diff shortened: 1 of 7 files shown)


zhut19:
Travis having trouble finding the new config files uploaded to strax_auxiliary_files yesterday. Now sure where the test tries to load the resources, does it also has a resource_cache somewhere?

petergaemers:
Sorry for not seeing this sooner.
The default config refers to the old one (fax_config.json) and not fax_config_nt.json

zhut19:
No worries, it's not in a hurry, last time I checked the strax in the default env on midway is still below 0.9.0. So I am still running tests. Hope to catch more issues before merging.

petergaemers:
Travis is now complaining about this:
```
    "Plugins without dependencies must return full strax "
460ValueError: Plugins without dependencies must return full strax Chunks, but RawRecordsFromFax produced a <class 'numpy.ndarray'>!
```

JelleAalbers:
Yes, you'll want to do something like 
```
return self.chunk(start=..., end=..., data=..., data_type=...)
```
see e.g. https://github.com/XENONnT/straxen/blob/master/straxen/plugins/daqreader.py#L299

petergaemers:
Finally, that took embarrasingly long. Lets merge this so we'll have a proper working version again

PR #43, opened 2020-04-22
Add field distortion effects w/ FDC map

zhut19:
Adding field distortion feature to core simulation.

Use SR1 nn tf part 1 for xenon1t
Return 0 correction for xenonnt

The results are shown below, observed positions with given points on a circle with r=[10, 20... 50]. The observed positions are stable after a few iterations, (as fdc map is not asking true position as input), more solid color represents later in the iteration (max=6)
image

Other fixes along with this pr:
Increase pax_interface truth buffer size,  because unlike strax, it's not being copied and cleared at the end of each chunk.

Check the length of instruction dtype names instead of instruction itself for extra columns, as now instruction is a structured array by default.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+git+git://github.com/AxFoundation/strax
 git+git://github.com/XENONnT/straxen
 uproot
 nestpy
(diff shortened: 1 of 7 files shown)


zhut19:
What happens is the instruction x, y, z of S2 is no longer the observed x, y, z. Instead, they should match the {x, y, z}_3d_nn values in the analysis. What's still we need to do later is see what's the best way to save those x, y, z in truth file.

PR #48, opened 2020-06-03
Seedopt

petergaemers:
This pull requests adds the option to set a seed for the random number generator.

### DIFF
--- a/wfsim/core.py
+++ b/wfsim/core.py
@@ -354,9 +354,8 @@ class S2(Pulse):
             positions = np.array([x_obs, y_obs]).T  # For map interpolation
         else:
             z_obs = z
-            print(x,y)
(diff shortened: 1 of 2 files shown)


zhut19:
Cool, it would also be great if this seed can also be used for all those np.random in the core.py

petergaemers:
Isn't setting the seed once enough?

According to this: https://stackoverflow.com/questions/59008887/what-is-the-working-range-of-numpy-random-seed
you set a global random state

zhut19:
I see, setting it once does do the job. Although, `rand_instructions` is not always called (at least my habit is writing instructions to csv files first). Does it make more sense to move the seed setting to the setup in FaxSimulatorPlugin?

petergaemers:
You are correct, let me fix that

PR #49, opened 2020-06-05
Adding features needed to support high energy channels for XENONnT simulations

petergaemers:
This pull request adds high energy and sum signals.

For this a change needed to be made in the plugins. Since high energy channels are not in XENON1T the plugins for 1T/nT are split off. So to switch between detectors register different plugins rather than setting the detector field as an option.

The sum signal is stored as part of the high energy channels, although it would in reality be part of the aquisition monitor

Note there is a not very pretty hard coded number of channels currently in https://github.com/XENONnT/WFSim/blob/455a481a44bc0ac6926f291e4e2148cb382c00dd/wfsim/core.py#L877

Usage:
```
strax.Context(
        storage=strax.DataDirectory('./strax_data'),
        register=wfsim.RawRecordsFromFaxNT,
        config=dict(detector='XENONnT',
                   fax_config='https://raw.githubusercontent.com/XENONnT/'
                 'strax_auxiliary_files/master/fax_files/fax_config_nt.json',
                    **straxen.contexts.xnt_common_config,),
        **straxen.contexts.common_opts)
```

I've made a pull request here to add these contexts to straxen
https://github.com/XENONnT/straxen/pull/127

I'll also update the tutorials

### DIFF
--- a/notebooks/Fax_Tutorial.ipynb
+++ b/notebooks/Fax_Tutorial.ipynb
@@ -48,8 +48,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "First we need to define the right context. The thing which differs now is where to get the plugin to provide raw records. By default this is the DAQ Reader. Now we do not want this so we register wfsim.RawRecordsFromFax. I think it is self explanatory where this plugin tells strax to get raw records from.\n",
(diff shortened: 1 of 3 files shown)


petergaemers:
Dear self,
Please stop using the web interface to change things, this never works.

PR #52, opened 2020-07-20
Update strax_interface.py

petergaemers:
This adds the nVeto simulator class


zhut19:
I guess this should be closed.

PR #53, opened 2020-07-31
Fix optical

mzks:
I would like to add methods for nVeto pulses. It strongly depends on the Geant4 optical simulation.
To prepare this implementation and test, I fixed obvious errors on 'optical', and add workaround lines to be fixed for the future.


zhut19:
Looks good to me, sorry about the long wait.

PR #54, opened 2020-10-09
Remove outdated per pmt pattern

zhut19:
The channel assignment using the pattern map is not fully utilized so far. It is time to remove the pattern function that mostly just for my personal usage, and back to pattern map interpolation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ with open('HISTORY.md') as file:
 
 setuptools.setup(
     name='wfsim',
-    version='0.0.4.1',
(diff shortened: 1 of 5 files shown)


zhut19:
In https://github.com/XENONnT/WFSim/pull/54/commits/b2d2696ad640bc0cfd35ef1ba7d11d3e1441a642 when PMT pattern map return zeros at the edge but still within tpc radius, the corresponding photon timings will be deleted, however, the truth will still record such an S2 peak, just with nan's in the photon timing fields.

PR #62, opened 2020-11-18
Below cathode

zhut19:
Allow below cathode S1 instructions to pass.


zhut19:
Also 'zle_threshold' and 'field_distortion_on' should be in fax_config and not in the strax plugin options.

PR #64, opened 2020-11-19
Remove read g4

zhut19:
The read g4 function can be in principle replaced by EPIX, however, it still would be nice to keep a read g4 function. This draft  PR is just trying to see removing the nestpy import can make the commits pass the Travis test again.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
 git+git://github.com/AxFoundation/strax
-git+git://github.com/XENONnT/straxen
 git+git://github.com/XENONnT/utilix
+git+git://github.com/XENONnT/straxen
 uproot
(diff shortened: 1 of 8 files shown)


zhut19:
@petergaemers hope you are okay with removing this read g4 part as epix is on the way. We can always add this back, but for now, nextpy is blocking the Travis check.

petergaemers:
Yes of course. This function served mainly as a placeholder for EPIX

jorana:
@zhut19 and @petergaemers , we solved the nestpy dep. in the base_env by downgrading to:
```nestpy==1.3.0     # WFsim dependency```
https://github.com/XENONnT/base_environment/pull/348

zhut19:
Thanks @jorana, but I guess I pulled the trigger too soon. So now it is in the hand of EPIX.

PR #67, opened 2020-11-24
add test for segmentation fault in noise handling

jorana:
Hi @petergaemers! In https://github.com/XENONnT/WFSim/pull/65 an unintended index error was introduced. It took me some time to figure out because with numba on it's harder to debug. It just died. Running with ``--debug`` only showed a segmentation fault.

When I found the problem, I actually spend most time on writing a test that should catch this in the future. Its a pain to write these things I admit. Hope it helps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,6 @@ python:
 - '3.7'
 install:
 - pip install -r requirements.txt
+- pip install hypothesis
 script:
 - python setup.py test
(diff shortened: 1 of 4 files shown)


petergaemers:
Hi Joran, thank you very much. I see there should have been a -1 in https://github.com/XENONnT/WFSim/blob/ac075ce1f0f6f8d1cedd2ffadac4cec89703ffbc/wfsim/core.py#L1132
But this also works

And thanks for the test, these things are horrible to write :P

jorana:
True, but also `ix >= len(data[ch])`, so there were two places where we could go out of range.

PR #68, opened 2020-12-03
Update for moved resource files

zhut19:
nT related resource files are being removed from `strax_auxiliary_files` See XENONnT/strax_auxiliary_files#9 and XENONnT/private_nt_strax_aux_files#1

PR #69, opened 2020-12-16
use files database

jorana:
Use https://straxen.readthedocs.io/en/latest/config_storage.html#downloading-xenonnt-files-from-the-database to get nT files.

For nT, we can now use (if we are on the latest straxen master) the database to get the required files. All the files from https://github.com/XENONnT/strax_auxiliary_files and https://github.com/XENONnT/private_nt_aux_files are uploaded to the database. We should be able to download those files like this. 

Some testing from the WFSim experts would be appreciated because they are most likely much better in isolating potential issues than I am.

### DIFF
--- a/tests/test_wfsim.py
+++ b/tests/test_wfsim.py
@@ -12,15 +12,20 @@ run_id = '010000'
 def test_sim_1T():
     """Test the 1T simulator (should always work with the publicly available files)"""
     with tempfile.TemporaryDirectory() as tempdir:
+        testing_config_1T = dict(
+            hev_gain_model=('to_pe_constant', 0.0085),
+            gain_model=('to_pe_constant', 0.0085)
+        )
         st = strax.Context(
             storage=tempdir,
             config=dict(
                 nchunk=1, event_rate=1, chunk_size=10,
                 detector='XENON1T',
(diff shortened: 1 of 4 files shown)


ramirezdiego:
Does this PR fully overwrite https://github.com/XENONnT/WFSim/pull/68?

jorana:
I must admit that I overlooked #68 , there are some misconceptions in that PR but also some useful solutions to other problems. Let me comment in that PR

jorana:
@zhut19 , can you fix the 1T configuration file? At this point there is little I can do and something that is actually broken on the WFSim side.

jorana:
@zhut19 , thanks for the update of the 1T config, stuff is still failing though. I'm quite sure you introduced some bug in https://github.com/XENONnT/WFSim/commit/c9ed8fbe16f70139bc01a499c77c869654ae6544. Please have a look. The `sc_gain` is empty for the 1T test.

I also added the option to test both 1T and nT in wfsim, these kind of errors would preferably not go unnoticed for such a long time. At the moment travis is limited to testing 1T because the nT files are private. To test `nT` you can `pytest` locally.

zhut19:
Thanks! @jorana

ramirezdiego:
Ready to merge?

jorana:
as far as I'm concerned but maybe @zhut19 or @petergaemers should do it

PR #71, opened 2021-01-18
Fix missing configs

jorana:
In https://github.com/XENONnT/private_nt_aux_files/pull/3/commits/35a1a79a009cf7d4c571dbb7eb95876eef423d82 some configs were deleted that were needed in WFSim, we can set them as defaults for nT.

I leave it to the developers to decide if they want to reinstate those entries in the `fax_config_nt.json` or just upload a channel map to the config or so with start and stops.

### DIFF
--- a/wfsim/core.py
+++ b/wfsim/core.py
@@ -298,7 +298,7 @@ class S1(Pulse):
         super().__call__()
 
     def photon_channels(self, points, n_photons):
-        channels = np.array(self.config['channels_in_detector']['tpc'])
+        channels = np.arange(len(self.config['gains']))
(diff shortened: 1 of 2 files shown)


petergaemers:
Hi Joran. Thanks for your effort. In my upcomming update I remove this part and read the channel map since that already have everything we need

jorana:
That indeed sounds like the approach that makes most sense. I needed to get it working with the current config so this just took me 2 minutes. Feel free to close at your convenience ;)

petergaemers:
Fixed in https://github.com/XENONnT/WFSim/pull/72

PR #72, opened 2021-01-27
Wfsim updates

petergaemers:
Work in progress
1. RawRecordsEpix has a bug in the chunk start/end time

This pull request adds:

- nVeto integration
- Epix integration
- strax_interface cleanup to make it nicer
- Enable/disable noise
- Small changes to handeling instructions. There was some issue reading/writing strings so the 'recoil' field is now an int
- Added fields to instructions for g4 input (e_dep, g4_id)
- Anode warping for S2s


wfsim/strax_interface.py
```diff
@@ -13,23 +13,28 @@
...
+             (('Quanta type (S1 photons or S2 electrons)', 'type'), np.int8),
+             (('Time of the interaction [ns]', 'time'), np.int64),
+             (('End time of the interaction [ns]', 'endtime'), np.int64),
+             (('X position of the cluster[cm]', 'x'), np.float32),
+             (('Y position of the cluster[cm]', 'y'), np.float32),
+             (('Z position of the cluster[cm]', 'z'), np.float32),
+             (('Number of quanta', 'amp'), np.int32),
+             (('Recoil type of interaction.', 'recoil'), np.int),
+             (('Energy deposit of interaction', 'e_dep'), np.float32),
+             (('Eventid like in geant4 output rootfile', 'g4id'), np.int32)
```
ramirezdiego, line 28: Cool, thanks. Can we also add [this guy](https://github.com/XENONnT/epix/blob/master/epix/io.py#L164)?

wfsim/strax_interface.py
```diff
@@ -60,34 +65,46 @@ def rand_instructions(c):
...
+        timings = e["pmthitTime"].array(library="np")*1e9
 
     ins = np.zeros(n_events, dtype=instruction_dtype)
 
     ins['x'], ins['y'], ins['z'], ins['time'] = xp.flatten()  / 10, \
-                                                       yp.flatten() / 10, \
-                                                       zp.flatten() / 10, \
-                                                       1e7 * time.flatten()
+        yp.flatten() / 10, \
+        zp.flatten() / 10, \
```
ramirezdiego, line 106: Duplicated division by 10 (see 20 lines above)?

wfsim/core.py
```diff
@@ -316,7 +318,7 @@ def photon_timings(self, t, n_photons, recoil_type):
             return
 
         if (self.config.get('s1_model_type') == 'simple' and 
-           recoil_type.lower() in ['er', 'nr']):
+           recoil_type in [1, 2]):
```
ramirezdiego, line 321: Does this now conflict with the string input from epix (`er`/`nr`)? Should we add more types, according to the input we give to NEST?

wfsim/strax_interface.py
```diff
@@ -60,34 +65,46 @@ def rand_instructions(c):
...
 
-    n_events = len(e.array('eventid'))
+def read_optical(file, c):
+
+    data = uproot.open(file)
+    if c['mc_version_above_4']:
+        e = data.get('events')
+    else:
+        all_ttrees = dict(data.allitems(filterclass=lambda cls: issubclass(cls, uproot.tree.TTreeMethods)))
+        e = all_ttrees[next(iter(all_ttrees))]
```
ramirezdiego, line 76: See these lines doing a similar job: https://github.com/XENONnT/epix/blob/master/epix/io.py#L91-L102 You wouldn't need the `mc_version_above_4` arg in that case.

wfsim/core.py
```diff
@@ -278,7 +278,7 @@ def __call__(self, instruction):
             # shape of recarr is a bit strange
             instruction = np.array([instruction])
 
-        _, _, t, x, y, z, n_photons, recoil_type, *rest = [
+        _, _, t, _, x, y, z, n_photons, recoil_type, *rest = [
```
WenzDaniel, line 281: Hey, would not it be easier to use here the array field-names? In this way you do not have to change always those lines when a new columns is added to the truth array. I know this will be conflicting with the pattern maps since the scipy.interpolated function cannot handle structured arrays for whatever reason, but you can circumvent this issue via: 

```
 pos = np.array([instructions['x'], 
                        instructions['y'], 
                        instructions['z']]).T
```

HenningSE:
This line https://github.com/XENONnT/WFSim/blob/wfsim_updates/wfsim/strax_interface.py#L111 needs to be adapted with the new integer recoil type I think.

wfsim/strax_interface.py
petergaemers: Sure

wfsim/strax_interface.py
petergaemers: Yeah this snuck in at some point

wfsim/core.py
petergaemers: No it is only used for s1 scintillation models (and curently ignored even). Furthermore my epix strax plugin imports the dtype from wfsim

wfsim/strax_interface.py
petergaemers: There is still a change in offset of nveto pmts

wfsim/core.py
petergaemers: Thanks, I'll need to check

wfsim/strax_interface.py
ramirezdiego: Uhm, but that version was in fact buggy due to the PMT DAQ naming issue, so this is probably a reason to just not use nVeto sims before MC v4. But okay, just stick to what @mzks and you agree on.

petergaemers:
@mzks @WenzDaniel Where can I find a small nveto input file to test if I didn't make mistakes?

wfsim/strax_interface.py
```diff
@@ -317,11 +288,28 @@ def __call__(self, instructions, channels, timings):
...
-                 default=('to_pe_per_run', 'to_pe_nt.npy'),
+                 default=('to_pe_per_run', 'https://github.com/XENONnT/private_nt_aux_files/blob/master/sim_files/to_pe_nt.npy?raw=true'),
                  help='PMT gain model. Specify as (model_type, model_config)'),
-    strax.Option('detector', default='XENONnT', track=True),)
+    strax.Option('detector', default='XENONnT', track=True),
+    strax.Option('channel_map', track=False, type=immutabledict,
+                 help="immutabledict mapping subdetector to (min, max) "
+                      "channel number. Provided by context"),
+    strax.Option('n_tpc_pmts', track=False,
+<<<<<<< HEAD
```
HenningSE, line 300: I tried to run this branch but I get an error for this line when importing wfsim.

wfsim/strax_interface.py
```diff
@@ -317,11 +288,28 @@ def __call__(self, instructions, channels, timings):
...
+                 help="Number of pmts in tpc. Provided by context"),
+    strax.Option('n_top_pmts', track=False,
+                 help="Number of pmts in top array. Provided by context"),
+=======
```
HenningSE, line 304: Think this must be removed as well.

wfsim/strax_interface.py
```diff
@@ -317,11 +288,28 @@ def __call__(self, instructions, channels, timings):
...
+                 help="immutabledict mapping subdetector to (min, max) "
+                      "channel number. Provided by context"),
+>>>>>>> aefdcb10a09bdc9cc4cd21fc277cfe29117f443a
```
HenningSE, line 307: And here

wfsim/strax_interface.py
```diff
@@ -422,7 +387,49 @@ def source_finished(self):
...
+        assert np.all(self.instructions['z'] < 0.25), \
+                "Interation is outside the TPC"
+        assert np.all(self.instructions['amp'] > 0), \
+                "Interaction has zero size"
+
+
+    def infer_dtype(self):
+        dtype = {data_type:strax.raw_record_dtype(samples_per_record=strax.DEFAULT_RECORD_LENGTH) 
+                for data_type in self.provides if data_type is not 'truth'}
+<<<<<<< HEAD
```
HenningSE, line 424: And here I think.

wfsim/strax_interface.py
```diff
@@ -422,7 +387,49 @@ def source_finished(self):
...
+        dtype['truth']=instruction_dtype + truth_extra_dtype
+=======
```
HenningSE, line 426: Same as the other comments.

wfsim/strax_interface.py
```diff
@@ -422,7 +387,49 @@ def source_finished(self):
...
+        dtype['truth'] = instruction_dtype + truth_extra_dtype
+>>>>>>> aefdcb10a09bdc9cc4cd21fc277cfe29117f443a
```
HenningSE, line 428: And here.

petergaemers: Forgot to merge this

wfsim/strax_interface.py
petergaemers: Oke then I'll change it to throw an exception when someone is using the old mc

wfsim/core.py
```diff
@@ -433,15 +435,19 @@ def inverse_field_distortion(self, x, y, z):
         positions = np.array([x_obs, y_obs]).T 
         return z_obs, positions
 
-    def luminescence_timings(self, shape):
+    def luminescence_timings(self, xy, shape):
```
zhut19, line 438: Oops, didn't realize this line sneaked in with my commit. It's a bit tricky here, xy should a 2d array of [n_inst., 2], where the `shape` is [n_inst., n_photon] (the xy doesn't need to be the same for all inst.). So from here, the gas gap and the following pulse model need to be calculated for each instruction.
(I can also take care of this if @petergaemers can share the gas gap map?)

petergaemers: Hi Tianyu, the gas gap map is on the private aux repo: https://github.com/XENONnT/private_nt_aux_files I did upload it to the wrong directory I just see

WenzDaniel:
@petergaemers I had hope @mzks would answer your question. I do not have any recent nVETO files.

mzks:
[mc_test_output.zip](https://github.com/XENONnT/WFSim/files/5894096/mc_test_output.zip)
Here you are.

petergaemers:
@mzks Awesome, thanks a lot

PR #76, opened 2021-02-04
Fix Garfield

petergaemers:
As pointed out by @WenzDaniel _luminescence_timings_index is recompiled every time by numba when defined in the function. (Or at least that what it looks like). This is what murders our speed. Now it should work as it's supposed to.

While I'm at it, lets also fix the tests here


zhut19:
I am actually fixing this already in #75. Should I merge that before this?

petergaemers:
Ah if thats the case we can forget about this

PR #79, opened 2021-02-05
Update to use nest ids

petergaemers:
Update to use nest ids to refer to S1 scintilation models, note only ER actually makes sense.
Also change detector field name to accomidate epix and placeholder functions to strax_interface


wfsim/strax_interface.py
```diff
@@ -49,7 +49,7 @@ def rand_instructions(c):
     instructions['event_number'] = np.digitize(instructions['time'],
          1e9 * np.arange(c['nchunk']) * c['chunk_size']) - 1
     instructions['type'] = np.tile([1, 2], n)
-    instructions['recoil'] = [1 for i in range(n * 2)]
+    instructions['recoil'] = [7 for i in range(n * 2)] #Use nest ids for  ER
```
ramirezdiego, line 52: Is it too late to comment here? There is no error, but let's just remember to document somewhere that the rand_instructions produce gammas only.

jorana: you can also just `from nestpy.helpers import NEST_INTERACTION_NUMBER` and be done with it :) 
https://github.com/NESTCollaboration/nestpy/blob/master/src/nestpy/helpers.py#L27

jorana: also it's written down here:
https://github.com/XENONnT/WFSim/blob/master/wfsim/core.py#L22

ramirezdiego: I don't understand what you mean in the second comment.

jorana: the way I phrased it not correct so I understand you don't understand 😄 . Never-mind, it wouldn't hurt to indeed explicitly add in the NestId what 7,8,11 are like you mention.

PR #80, opened 2021-02-05
Return correct PMT ID for nVeto

mzks:
WFSim uses PMT ID started on 0 for also nVeto option.
However, it is convenient to return the correct ID (2000--2119) is useful for analyst.
I've checked it works, please review the lines.


zhut19 (approved):
Looks good with me, thanks @mzks

PR #81, opened 2021-02-05
Geant4 event ID in truth

mzks:
I added 'g4id' in truth using optical simulation. (In the current implementation, they are empty.)
It works.

In the far team-D telecon, I said unique event ID is enough to match TPC and nVeto events, but it is not correct.
Multi 'raw_records' and 'raw_records_nv' are created in one Geant4 event. In other words, more raw records are generated for one truth. We cannot match that because we don't know which truth created one specific raw record.
(Using event time, we can guess that but I think it is not straight-forward solution.)
Am I missing important thing? Please give me a comment and let's start a discussion.

### DIFF
--- a/wfsim/strax_interface.py
+++ b/wfsim/strax_interface.py
@@ -92,7 +92,7 @@ def read_optical(c):
     ins['z'] = e["zp_pri"].array(library="np").flatten() / 10.
     ins['time']= 1e7 * time.flatten()
     ins['event_number'] = np.arange(n_events)
-    ins['g4id'] = np.arange(event_id)
+    ins['g4id'] = event_id
     ins['type'] = np.repeat(1, n_events)
     ins['recoil'] = np.repeat(1, n_events)
     ins['amp'] = [len(t) for t in timings]
--- a/wfsim/strax_interface.py
+++ b/wfsim/strax_interface.py
@@ -71,7 +71,8 @@ def read_optical(c):
     except:
         raise Exception("Are you using mc version >4?")
 
-    n_events = len(e['eventid'].array(library="np"))
+    event_id = e['eventid'].array(library="np")
+    n_events = len(event_id)
     # lets separate the events in time by a constant time difference
     time = np.arange(1, n_events+1)
 
@@ -91,6 +92,7 @@ def read_optical(c):
     ins['z'] = e["zp_pri"].array(library="np").flatten() / 10.
     ins['time']= 1e7 * time.flatten()
     ins['event_number'] = np.arange(n_events)
+    ins['g4id'] = np.arange(event_id)
     ins['type'] = np.repeat(1, n_events)
     ins['recoil'] = np.repeat(1, n_events)
     ins['amp'] = [len(t) for t in timings]


WenzDaniel:
Hej Keita, thanks a lot. I let Peter or Tianyu approve the PR itself. I do not fully understand your point about the eventid and the matching between TPC and nVETO data. I think for matching those two detectors the event-id should be sufficient. However, on the TPC side I agree that one would need both an event- and an interaction-id for a very precise matching. 
One additional thing I noticed and which we should fix. Currently, I see a big issue with your "time"-fields:

Since your are not using any time information from GEANT4 your loosing basically everything is not it? You do not know any relative arrival times between photons, in fact it is even worse your photons will be well separated by 10**7 ns which means none of your photons will pass our software trigger. Further, you lose also all relative information with the TPC.

mzks:
Dear Daniel, thank you very much for your comments. I may misunderstand the concept to match TPC and nV records.
I'm confusing how to match these events.
First, Geant4 make an event with specific g4id for TPC and nVeto. These ids are passed to WFSim finally, but it saved in `truth`.  The `truth` is generated by each original event, thus there are many records for one truth. How can we find the original G4 ID for a `records`?

mzks:
https://github.com/XENONnT/WFSim/blob/65487a570d8746bf57a590a04e82996e2426e40f/wfsim/strax_interface.py#L81
Is this time information by Geant4?
This is relative timing in one Geant4 event. WFSim will generate events in a static timing gap with the relative timing.

WenzDaniel:
Hej Keita, could you explain me than the difference between `timings` and `time` in `ins`?

mzks:
>> Since for the TPC EPIX generates the timestamps for the instructions a matching may become very tricky.
That's problem. I don't fully agree to go EPIX with nVeto immediately, but we need to make consistent timestamp for both finally.

mzks:
As you know, Geant4 has global time in each event, whose t=0 is timing to generate initial particle(s). Then, particles and opticalphotons are transported and recorded in some timing. The timing is `timing` in Geant4 output.

The tool expands the timing, assuming that each event occurs at evenly spaced times.
In other words, `time` is the offset for converting each event to a time stamp.
In EPIX, it seems that this offset is randomly determined. We need to do the same with nVeto.
On the other hand, I don't immediately agree with using EPIX with nVeto, because I want to avoid mass production of tools and I don't know if we should make another set of data that has almost the same meaning (and size) as the output of Geant4.
https://github.com/XENONnT/epix/blob/de75d6ec95a08c7e5b6a534ebddb46a6a2dda0e9/epix/event_separation.py#L43

WenzDaniel:
> On the other hand, I don't immediately agree with using EPIX with nVeto, because I want to avoid mass production of tools and I don't know if we should make another set of data that has almost the same meaning (and size) as the output of Geant4.

I am open for discussion. But for the sake of robustness and to minimize mistakes, I think we should use epix. The addition would be rather small.

petergaemers:
Hi Keita, thank you very much.

The age old question of "where to up the QEs" is a never answered problem. I think the best place to put it is in the read_optical function. Like that the nveto and Andrii's optical mc module uses it. Epix is not a very useful place since it's not used for optical mc.

ramirezdiego:
Hey Peter, I also think that epix does not make sense a priori, since it isn't tailored for optical sims indeed, but then let's discuss on Wednesday how to handle the timing information. Currently we feed WFsim with events separated by some offset (although all events start at 0 in G4), in which the t0 of the event is such of the first interaction in the TPC. We would need to apply that correction to the nVeto timings. And even that might be insufficient, as we could get negative times for the nVeto interactions that happen earlier than these in the TPC. Can be tricky...

@mzks Remember to drop the csv with the nVeto PMTs QEs in the auxiliary repo: https://github.com/XENONnT/private_nt_aux_files

mzks:
Thank you very much for fruitful discussions, everyone. In particular, @WenzDaniel !
Let's discuss about that on the next telecon.

PR #84, opened 2021-02-08
add h-diffusion function

zhut19:
We have been ignoring horizontal diffusion so far, which might impact the S2 pattern.

The added function in S2 pattern generation would now output diffused patterns, but with naive calculations

  - One can set an additional parameter `diffusion_constant_liquid_horizontal` in config to enable this
  - We will create updated xy positions for all electrons in each S2, the new pattern is calculated averaging over those new positions.
    - Positions outside the TPC radius will be ignored, as the output will be the average over all other positions.
    - It's not a weighted average, as an electron with 50 photon carries the same weight as an electron with 20 photon
    - There's no enforced time-positon correlation at this per electron level.

### DIFF
--- a/wfsim/core.py
+++ b/wfsim/core.py
@@ -619,12 +619,12 @@ class S2(Pulse):
     def s2_pattern_map_hdiff(self, n_electron, z, xy):
         drift_time_gate = self.config['drift_time_gate']
         drift_velocity_liquid = self.config['drift_velocity_liquid']
-        diffusion_constant_liquid_horizontal = getattr(self.config, 'diffusion_constant_liquid_horizontal', 0)
+        diffusion_constant_transverse = getattr(self.config, 'diffusion_constant_transverse', 0)
 
         assert all(z < 0), 'All S2 in liquid should have z < 0'
 
         drift_time_mean = - z / drift_velocity_liquid + drift_time_gate  # Add gate time for consistancy?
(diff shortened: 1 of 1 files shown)


zhut19:
It seems to be running fine

petergaemers (approved):

terliuk:
Hi, I have a few comments:

1. How do we plan to control whether diffusion is applied? Do we just set diffusion to 0? (will it break in this case?)
2. Is it possible to set/overrride it using configuration and not editing fax file? 
3. (warning, taste opinion, we can ignore it for now) I don't like "diffusion_constant_liquid_horizontal" , prefer "diffusion_constant_transverse" and "diffusion_constant_longitudinal", as this is how it's referred in most of the literature.

petergaemers:
Adding an "enable_horizontal_diffusion" flag to the config is a good idea

ramirezdiego:
@zhut19 Since we're here, could you add docstrings for `s2_pattern_map_hdiff` and `photon_channels`?

zhut19:
Thanks for the comments!
@terliuk and @petergaemers I hope this `getattr(self.config, 'diffusion_constant_transverse', 0) > 0` does the trick already. Both 0 or None will bypass the added function. A flag feels a bit redundant sometimes unless we try to track it.
@terliuk probably you already know this now, the `fax_config_override` does the trick.
@ramirezdiego you got it.

PR #83, opened 2021-02-08
S1 refactor + Readout bug fix

petergaemers:
This pull request:

- Makes all s1 functions callable from the outside. So now they're all staticmethod so if you want to calculate some photon arrival timings yourself this is now easy.
- Fix readout bug. The last part of the data was not read out from the buffer.
- Make raw_optical slightly nicer

Side remark: Tests actually work when run locally. Through github we have issues with the private repo

PR #87, opened 2021-02-10
No lower case for recoil_type

ramirezdiego:
`recoil_type` is now an integer.


petergaemers:
Fixed in https://github.com/XENONnT/WFSim/pull/83

ramirezdiego:
Ah, perfect!

PR #88, opened 2021-02-10
reinstate tests

jorana:
Reinstate some tests. 

Perhaps from breaking this again in the future, I think adding a default config would help. For example, the xenon1t fax file is missing some configs, see the TODOs. It should be doable to test without access to the private files, most things are just a bunch of numbers in the end (ok, some pattern maps etc complicate things a bit).

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,6 @@
 [bumpversion]
-current_version = 0.1.0
+current_version = 0.1.1
 files = setup.py wfsim/__init__.py
 commit = True
 tag = True
+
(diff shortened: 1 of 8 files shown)


petergaemers:
Thanks Joran

jorana:
OK, #83 broke the tests again...

jorana:
ok so apart from the TODO's the test is now really failing due to errors induced in #83, all the more showing the need for testing

petergaemers:
Should be fixed in https://github.com/XENONnT/WFSim/pull/91

PR #94, opened 2021-02-15
Correct detector naming

ramirezdiego:
Our old `"xenonnt_detector"` was reintroduced in here: https://github.com/XENONnT/WFSim/pull/83
@jorana @petergaemers Shouldn't the test have triggered something in this case, since `ly` isn't called?

I will push a new tag after this is merge, if you agree.

### DIFF
--- a/wfsim/core.py
+++ b/wfsim/core.py
@@ -308,7 +308,7 @@ class S1(Pulse):
 
     @staticmethod
     def get_n_photons(n_photons,positions, s1_light_yield_map, config):
-        if config['detector']=='xenonnt_detector':
+        if config['detector']=='XENONnT':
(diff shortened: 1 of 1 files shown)


petergaemers:
Hi Diego, thanks a lot.
Tests pass because this guy:
https://github.com/XENONnT/WFSim/blob/57d1ce6a5dd33a613755b8314b1f62c8caf7aaa9/tests/test_wfsim.py#L47

jorana:
Also, the test at the moment does not do anything for nT:
https://github.com/XENONnT/WFSim/blob/57d1ce6a5dd33a613755b8314b1f62c8caf7aaa9/tests/test_wfsim.py#L38

PR #95, opened 2021-02-15
Fix get_n_photons

HenningSE:
Looks like we missed to change `self.config` to `config` in the previous PRs https://github.com/XENONnT/WFSim/pull/94 and https://github.com/XENONnT/WFSim/pull/83.

### DIFF
--- a/wfsim/core.py
+++ b/wfsim/core.py
@@ -310,7 +310,7 @@ class S1(Pulse):
     def get_n_photons(n_photons,positions, s1_light_yield_map, config):
         if config['detector']=='XENONnT':
             ly = np.squeeze(s1_light_yield_map(positions),
-                            axis=-1)/(1+self.config['p_double_pe_emision'])
+                            axis=-1)/(1+config['p_double_pe_emision'])
         elif config['detector']=='XENON1T':
             ly = s1_light_yield_map(positions)
             ly *= config['s1_detection_efficiency']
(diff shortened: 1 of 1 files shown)


HenningSE:
Trying to run this branch in a notebook on dali. Looks like there is at least one other problem we/I have to fix before it works fine.

HenningSE:
Got this error message linking to https://github.com/XENONnT/WFSim/blob/Fix_get_n_photons/wfsim/core.py#L59-L61. 
WFsim_error_NoneType
Don't know at the moment how to solve it.

petergaemers:
Hi Henning can you explain what you're trying to do?
Because this is not related to https://github.com/XENONnT/WFSim/pull/94 or https://github.com/XENONnT/WFSim/pull/83

petergaemers:
Hi Henning, thanks.
The error you're getting is that you're trying to simulate 0 photons. What are you doing to get this?

HenningSE:
I'm using different instructions files generated with EPIX and then run WFsim. Checked the instructions file and `amp` is always larger 0. Could this be something like afterpulses?

HenningSE:
With the last commit i changed the return of `photon_timings` in case of 0 photons from nothing to an empty numpy array.  Now the previously reported error is gone and the simulation seems to work fine. I'm not 100% that this "fix" is right so it would be great if someone more familiar with the code can check. 

I think we did not have this problem before since a empty array was set: `self._photon_timings = np.array([])` in S1 `__call__`. (Don't know how to properly link it so here is a screenshot from https://github.com/XENONnT/WFSim/pull/83:
WFsim_PR_missing_empty_array

PR #96, opened 2021-02-16
Fix 1t config for test

petergaemers:
Fix the commit id of the 1t config.


zhut19:
Right, thanks Peter, will remember to change that next time.

PR #98, opened 2021-02-17
Change TTS from sigma to FWHM

terliuk:
PMT datasheets give TTS as FWHM, so sigma has to be divided by 2.35482 in order to get sigma for Gaussian.


terliuk:
Any opinion about scaling FWHM to sigma like I did here? Or is it better to change definition of the value in our config files?

skazama (approved):
Thanks Andrii. I think it does not matter as long as it is properly documented: in the config file, it should be mentioned that  9 ns TTS value is not sigma but FWHM.

zhut19 (approved):

Issue #82, opened 2021-02-05
How to implement PMT QE for nVeto


mzks:
I create very dirty, broken implementation to add PMT QE for nVeto PMTs. (branch `nveto_pmt_qe`).
The implementation is bad, and it generates errors.

In my naive prediction, only what we have to do is dropping channels and timings after reading Geant4 input, but I may care about ins (instruction) at the same time.
Could you please give me a hint to implement?

For your information,
For nVeto PMTs, I will add static number of QE(*CE) for all PMTs because there are no QE specification sheet for all PMTs. (Half PMTs have measured values.)

WenzDaniel:
Hi Keita, also with respect to the time-field issues I mentioned in your PR, I think the best solution here would be to let EPIX handle photon timing and QE for the Veto systems. In this way EPIX would generate two instructions one for the TPC and one for each Veto. In this way we can ensure a correct treatment of all timestamps.

ramirezdiego:
Hi Keita, I think I responded here to one of your questions, sorry for the mess: https://github.com/XENONnT/WFSim/pull/81

pietrodigangi:
Hi @mzks, just to add that, besides where, I think we should model the nVeto PMT's QE as best as available info allow us.
We have QE spectra for 63 installed PMTs out of 120 in this note: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:dsg:muonveto:small_wt:pmts:quantumefficiencybologna
Unfortunatelly Hamamatsu only provided them in their pdf datasheet. with some patience we can extract the spectra from their tables. and randomly assign a QE spectrum to each of the remaining 57 PMTs, taking it from the available ones.

mzks:
I've already extracted the QE for 65 available PMTs from pdf.
![QElist_all](https://user-images.githubusercontent.com/12980386/107396089-c90bba00-6b40-11eb-824e-ec86e823bc68.png)
The solution which the tables use for 65 PMTs and {random QEs or  average QEs} for the other PMTs, is not beautiful. In addition, the current WFSim doesn't load `pmthitenergy` in Geant4 to calculate them. I think that we can add just an effective constant as a typical QE in this term.

pietrodigangi:
wonderful that you already extracted those spectra! :)
[...]

PR #99, opened 2021-02-21
neutron Veto PMT QE implementatiion

mzks:
I added nVeto PMT QE as a one possible solution for #82 .
I implemented the QE using provided QEs depends on wavelengths of opticalphotons.
I believe it is the current best way.
The implementation is naive, like C++ code but it is not bottle-neck of the entire simulator.


wfsim/strax_interface.py
```diff
@@ -289,6 +320,10 @@ def __init__(self, config):
                  help="Number of pmts in top array. Provided by context"),
     strax.Option('neutron_veto', default=False, track=True,
                  help="Flag for nVeto optical simulation instead of TPC"),
+    strax.Option('nv_pmt_qe_file', default='https://github.com/XENONnT/private_nt_aux_files/blob/master/sim_files/nveto_pmt_qe.json', track=True,
+                 help="path and filename to nveto PMT QE json file"),
+    strax.Option('nv_pmt_ce_factor', default=1.0, track=True,
+                 help="Collection efficiency factor for neutron Veto PMT"),
```
ramirezdiego, line 326: These are/will be fixed values, so no chance to test nVeto under different QE/CE conditions. @zhut19 @petergaemers what's the policy here? Should the CE value simply be added to `fax_config`?

petergaemers: This should go in the config. This is for plugin configuration. Simulation config should go in the config file

ramirezdiego: Shouldn't the QE file go in `load_resource.py`? Sorry if I miss something obvious.

mzks: The reason why I load them not in `load_resource.py` is I would like to drop optical hits in `read_optical()`. In there, `Resource` class is not available, is it right?

mzks: I moved them to private_nt_aux_files/sim_files/fax_config_nt.json.
This implementation works for ~6,000 events MC input in dali notebook and my local environment.

mzks:
I changed to load the resource in `load_resouce` following Diego's suggestion.

wfsim/strax_interface.py
petergaemers: Keita, you are correct :)

mzks:
And I'm sorry this PR includes some general bug fix for nVeto optical simualtions.

petergaemers:
Thank you very much Keita. Can you  check what CodeFactor and travis are complaining about?

mzks:
Done! Please give me another comment or merge it.

jorana (commented):
@mzks , could you add a docstring to `read_optical_nveto_with_cut`? Such like how the config (c) and the event (e) should roughly look like and which steps are taken in the `read_optical_nveto_with_cut` function? 

It might be a bit too much to ask if we can have it tested. In the future however, it might not be a bad idea to upload a dummy file to simulate on and test that.

petergaemers (approved):

wfsim/strax_interface.py
```diff
@@ -78,8 +80,40 @@ def read_optical(c):
 
     if c['neutron_veto']:
         nV_pmt_id_offset = 2000
-        channels = [[channel - nV_pmt_id_offset for channel in array if channel >=2000] for array in e["pmthitID"].array(library="np")]
+        channels = [[channel - nV_pmt_id_offset for channel in array] for array in e["pmthitID"].array(library="np")]
         timings = e["pmthitTime"].array(library="np")*1e9
+        wavelengths = [[1239.84 / wavelength for wavelength in array] for array in e["pmthitEnergy"].array(library="np")]
```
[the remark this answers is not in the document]
ramirezdiego: This is hc in units of eV, to convert the energy in the PMTs into wavelengths, to pass to the QE correction. @mzks even if this PR was merged in a hurry, could you address this comment by Peter and also the one by @jorana on the dosctrings, in a new patch PR?

It would also be good if you can link your nVeto WFSim run example here, so that we can bless it and upload it to the readthedocs tutorial.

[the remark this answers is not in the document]
mzks: Thank you very much @ramirezdiego . I will do (sorry not today...)

PR #100, opened 2021-02-22
Make release

jorana:
Get ready for a new release that from now on is tracked [also in pipy](https://pypi.org/project/wfsim/). For the future, it would be great if we add a comment to the history such that we know which update was done when.


ramirezdiego (approved):
Yes, please. Let's make releases more often and start linking every single merged PR in the HISTORY.

PR #102, opened 2021-02-23
Update nT pattern maps

l-althueser:
This should follow to the PR https://github.com/XENONnT/private_nt_aux_files/pull/33.

For S2 map please revisit all appearances of the S2 pattern and ly map since it is no longer normalized! You can see https://github.com/XENONnT/private_nt_aux_files/pull/33 if you want to see how to open the map and how it looks like summed over all PMTs.

Don't merge this right now and add map related changes please. Thanks!


zhut19 (approved):
The map is good to use with the current code. We can merge here once https://github.com/XENONnT/private_nt_aux_files/pull/33 is merged.

PR #103, opened 2021-02-26
S2 refactor, Full chain simulator and many many docstrings

petergaemers:
This Pull Request:

- Adds wrapper class for full chain simulations.
- Refactors S2 class to have functions callable externally
- Adds many doc strings

Minimal working example of the full chain simulator is here:
https://github.com/XENONnT/mc_chain/blob/wfsim_update/run_WFsim.py

For running locally I needed to comment lines 69 throough 93 of load_resource.py to read the configs correctly
I used a not very recent strax_interface branch of epix, so it's possible there are already some things wrong/not working nicely anymore (like the source_rate config parameter)
You'll need to specify the fax and epix configs and a root file yourself.

Another small fix is the endtime is now the end of the last generated pulse
This should be all :)


petergaemers:
I'm not very sure if adding ntauxfiles as a requirment is very helpfull since thats a private repo. But it's needed to run load_resource

wfsim/strax_interface.py
```diff
@@ -100,31 +109,39 @@ def read_optical(c):
...
-    timings = e["pmthitTime"].array(library="np")*1e9
     constant_hc = 1239.841984 # (eV*nm) to calculate (wavelength lambda) = h * c / energy
-    wavelengths = [[constant_hc / energy for energy in array] for array in e["pmthitEnergy"].array(library="np")]
-
-    collection_efficiency = c['nv_pmt_ce_factor']
-    resource = load_config(c)
+    channels = [[channel - nV_pmt_id_offset for channel in array] for array in events["pmthitID"].array(library="np")]
+    timings = events["pmthitTime"].array(library="np")*1e9
+    wavelengths = [[constant_hc / energy for energy in array] for array in events["pmthitEnergy"].array(library="np")]
+    collection_efficiency = config['nv_pmt_ce_factor']
```
WenzDaniel, line 130: Out of curiosity, what kind od collection efficiency is this? I though QE would be enough.

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+    def set_timing(self,):
+        '''Set timing information in such a way to synchronize instructions for the TPC and nVeto'''
+        logging.info("Setting timings")
+        rate = self.context.config['event_rate']
+        start = self.context.config['event_start']
+        stop=self.context.config['event_stop']
+        n_events = stop-start
+
+        timings= (self.context.config['event_start']/rate + \
+                 wfsim.units.s*np.random.random(n_events)*(n_events)/rate).astype(np.int64)
```
WenzDaniel, line 253: I think here you have to change the logic slightly as we do not write empty events/instructions. 

E.g. if you simulate 15 k events you may only see 8k in the TPC simply due to the solid angle. Hence the timing must be computed like this 
```
np.random.uniform(first_g4_id/rate, last_g4_id/rate, n_events)
```
If "rate" should represent the decay rate of the source and not the event rate in the TPC.

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+        timings.sort()
+
+        #For tpc we have multiple instructions per g4id. For nveto we only have one
+        counter=Counter(self.instructions_epix['g4id'])
+        i_per_id = [counter[k] for k in range(start,stop)]
+        timings_tpc=np.repeat(timings,i_per_id)
+        self.instructions_epix['time']+=timings_tpc
+        if self.context.config['neutron_veto']:
+            self.instructions_nveto['time']=timings[self.instructions_nveto['g4id']]
```
WenzDaniel, line 262: See above

wfsim/strax_interface.py
WenzDaniel: There is one more thing which is not considered yet I think. You can have g4_id_tpc != g4_id_nveto. Hence you have to do something like  first_g4_id = min(g4_id_tpc, g4_id_nveto) and same for max.

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+    
+    def set_configuration(self,):
+        '''Set chunking configuration and feeds instructions to wfsim'''
+        max_events_per_chunk=500#is this hard coding really nessecairy...
```
WenzDaniel, line 266: You could add this number as an untracked option. I think we have to play here a bit to find the optimal data through put.

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+        chunk_size = np.ceil(max_events_per_chunk/self.context.config['event_rate'])
```
WenzDaniel, line 267: Why do we need to know this apriori? I do not excaty know how the chunking works in WFSim. Is data which exceeds the current chunk postponed into the next one? Would not this otherwise lead to one of the following problems: 

* data overlapping into the next chunk
* data cut at chunk bounds

 I thought we would set the chunk start according to the end of the previous chunk.

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+        nchunk = np.ceil((self.context.config['event_stop']-self.context.config['event_start'])/ \
```
WenzDaniel, line 268: Is this the nth chunk or n_chunks as in how many chunks we expect to get?

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+    def run_strax(self,run_id):
+        '''Runs wfsim up to raw records for tpc and if requisted the nveto.
+            All this setting neutron_veto=True/False is needed to get wfsim to do the right thing
+        '''
+        self.context.set_config(dict(neutron_veto=False))
+        self.context.make(run_id,'raw_records')
+
+        #nveto doesn't have proper afterpulses or noise implementation
+        #Cannot use neutron_veto=True/False here
+        if len(self.targets)>1:
```
WenzDaniel, line 296: This means in case of nveto stand alone simulations the user has to use a different tool right?

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+            self.context.make(run_id,'raw_records_nv')
+        
+    def _sync_meta(self,):
+        '''Sync metadata between tpc and nveto. The start and end times of the metadata must match for
+            something strax wants. Check which ever one starts first, make that the start of both. Same for end'''
+        logging.debug('Syncing metadata')
+        keys = ('start','end')
+        for key in keys:
+            if self.tpc_meta[key]>self.nveto_meta[key]:
+                self.tpc_meta[key]=self.nveto_meta[key]
```
WenzDaniel, line 307: This only works for the start as the start is given by min(tpc, nveto), but the end is given by max(tpc, nveto).

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+            else:
+                self.nveto_meta[key]=self.tpc_meta[key]
```
WenzDaniel, line 309: Also be careful you also have to updated the chunk start of the first chunk and the chunk end of the last chunk.

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+    
+    def sync_meta(self,run_id):
+        '''Sync metadata outputs from the tpc and the nveto.'''
+        self.tpc_meta=self.context.get_meta(run_id,'raw_records')
+        self.nveto_meta=self.context.get_meta(run_id,'raw_records_nv')
+        self._sync_meta()
+        self._store_meta()
+
+    def run_chain(self):
+        '''Main function. Sets instructions needed, runs wfsim and synch metadata'''
```
WenzDaniel, line 332: This guy is so little and cute and does so many things :D

wfsim/strax_interface.py
```diff
@@ -307,11 +486,11 @@ def __init__(self, config):
...
-    strax.Option('event_rate', default=5, track=False,
+    strax.Option('event_rate', default=1000, track=False,
                  help="Average number of events per second"),
     strax.Option('chunk_size', default=100, track=False,
                  help="Duration of each chunk in seconds"),
-    strax.Option('nchunk', default=1, track=False,
+    strax.Option('nchunk', default=10, track=False,
                  help="Number of chunks to simulate"),
     strax.Option('right_raw_extension', default=50000),
     strax.Option('timeout', default=1800,
```
WenzDaniel, line 496: I think timeout does not have to be tracked

WenzDaniel (commented):
Hej Peter thanks a lot for this gigantic PR. Took me almost 2 h to go through it. The doc strings and code cleaning will help a lot to understand the code better.

Many of the technical comments are about the nveto optical implementation and so @mzks might want to have a look too. I think one can make the code much faster and simpler if awkward arrays are used.

I also put some comments regarding the timing and chunking as I think it wont work yet, or I may have understood some parts wrongly. Although I think the changes I mentioned can be all considered as minor and quickly to be implemented if needed. 

I also had one larger suggestions regarding the 1T/nT map business. I do not know if it is feasible as I do not know much about the resource loading step.  Anyhow I just dropped the idea here.

wfsim/strax_interface.py
ramirezdiego: CE at the 1st dynode, which comes after (and isn't included) in the QE.

jorana (commented):
> 
> 
> I'm not very sure if adding ntauxfiles as a requirment is very helpfull since thats a private repo. But it's needed to run load_resource

This should not be required unless you need some files that are not merged yet from the private repo. If that is the case, all the more reason to wrap up https://github.com/XENONnT/private_nt_aux_files/pull/24.

Adding this as a requirement is indeed almost impossible to do right given the private nature of the repo.

wfsim/strax_interface.py
WenzDaniel: Interesting, how do you measure/estimate this without automatically including it in the QE?

ramirezdiego: We need a Japanese for this 😉 
cc @mzks

You can find some good references for the (TPC) PMTs characterization in this note: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:mc:g1_nt_projection

petergaemers:
> > I'm not very sure if adding ntauxfiles as a requirment is very helpfull since thats a private repo. But it's needed to run load_resource
> 
> This should not be required unless you need some files that are not merged yet from the private repo. If that is the case, all the more reason to wrap up [XENONnT/private_nt_aux_files#24](https://github.com/XENONnT/private_nt_aux_files/pull/24).
> 
> Adding this as a requirement is indeed almost impossible to do right given the private nature of the repo.

Yeah I did that to make it easier to run locally, but that should end up here

wfsim/strax_interface.py
WenzDaniel: Okay I see it is a relative value.  Thanls

mzks: Please let me explain a bit more about this part.
Tha PMT acceptance is generally considered as Q.E. (lambda, theta, phi) * C.E.(theta, phi) in each PMT. The only way to have them is just measurement, but we didn't. They are measured as a coupled value. I guess that Q.E. is not enough, but we don't know the entire value even if we considered optical behavior on our MC. I would like to keep tuning parameter at here as 'ce' for our future. However, it is super-fine tuning thus I afraid to make some bias at this part. I would like to use the value as 1.0 now, but we need to discuss some days.

wfsim/core.py
```diff
@@ -334,7 +363,15 @@ def photon_channels(points, n_photons, config, s1_pattern_map):
...
 
     @staticmethod
-    def photon_timings(t, n_photons, recoil_type, config):
+    def photon_timings(t, n_photons, recoil_type, config, phase):
+        '''Calculate distribution of photon arrival timnigs
+        :param t: 1d array of ints
+        :param n_photons: 1d array of ints
+        :param recoil_type: 1d array of ints
+        :param config: dict wfsim config
+        :param ohase: str "liquid"
```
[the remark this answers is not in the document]
petergaemers: This is to distinguish the xenon triplet lifetime from s1(liquid) to s2(gas)

wfsim/core.py
```diff
@@ -423,40 +483,87 @@ def __call__(self, instruction):
...
+    def inverse_field_distortion(x, y, z, resource):
+        '''For 1T the pattern map is a data driven one so we need to reverse engineer field distortion into the simulated positions
+        :param x: 1d array of float
+        :param y: 1d array of float
+        :param z: 1d array of float
+        :param resource: instance of resource class
+        returns z: 1d array, postions 2d array 
+        '''
         positions = np.array([x, y, z]).T
         for i_iter in range(6):  # 6 iterations seems to work
```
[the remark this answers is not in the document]
petergaemers: This is from @zhut19 . My guess it it's brute forced that 6 is good enough

wfsim/load_resource.py
```diff
@@ -53,19 +53,25 @@ def __init__(self, config=None):
...
             raise ValueError(f"Unsupported detector {config['detector']}")
 
         for k in set(config).intersection(files):
             files[k] = config[k]  # Allowing user to replace default with specified files
         commit = 'master'   # Replace this by a commit hash if you feel solid and responsible
-        url_base = f'https://raw.githubusercontent.com/XENONnT/strax_auxiliary_files/{commit}/sim_files'
+        if config['detector'] == "XENON1T":
+            url_base = f'https://raw.githubusercontent.com/XENONnT/strax_auxiliary_files/{commit}/sim_files'
+        if config['detector'] == "XENONnT":
+            url_base = f'/Users/petergaemers/Desktop/python/private_nt_aux_files/sim_files'
```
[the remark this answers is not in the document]
petergaemers: Load_resource is a continious pain. I load local files for which this is needed. This needs some general improvements

wfsim/load_resource.py
```diff
@@ -109,28 +114,30 @@ def __init__(self, config=None):
             lymap.data['map'] = np.sum(lymap.data['map'][:][:], axis=2, keepdims=True)
             lymap.__init__(lymap.data)
             self.s2_light_yield_map = lymap
-            self.s2_luminescence = straxen.get_resource(files['s2_luminescence'], fmt='pkl.gz')
+            self.s2_luminescence = ntauxfiles.get_sim_file(files['s2_luminescence'], fmt='pkl.gz')
             self.fdc_3d = dummy_map(result=0)
-            gas_gap_map = straxen.get_resource(files['gas_gap_map'], fmt='pkl')
+            gas_gap_map = ntauxfiles.get_sim_file(files['gas_gap_map'], fmt='pkl')
             self.gas_gap_length = lambda positions: gas_gap_map.lookup(*positions.T)
```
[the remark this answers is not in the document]
petergaemers: Same as above. Load_resource deserved a bit more love to work nicely

wfsim/strax_interface.py
```diff
@@ -79,7 +88,7 @@ def read_optical(c):
     time = np.arange(1, n_events+1)
 
     if c['neutron_veto']:
```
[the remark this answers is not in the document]
petergaemers: You mean QEs for both detectors? Won't be hard, someone will need to do it. There is some ongoing discussion every now and then where the tpc qes should go in that case

wfsim/strax_interface.py
```diff
@@ -100,31 +109,39 @@ def read_optical(c):
...
+    for ievent, _ in enumerate(channels):
         new_channels = []
         new_timings = []
         for j, _ in enumerate(channels[ievent]):
             rand = np.random.rand() * 100
             channel = channels[ievent][j]
+            #This trows away accidental none-nveto hits (so in tpc or mv)
             if not (0 <= channel < 120):
                 continue
             if not (timings[ievent][j] < 1.e6): # 1 msec. as a tentative
```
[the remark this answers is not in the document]
petergaemers: Sometimes a photon hits the muon veto

wfsim/strax_interface.py
petergaemers: Is the first comment refering to the offset part?

For the second part. If the ids are not the same you cannot synchronize at all
Or do you mean the first g4 id in tpc in not nessecairy the first in nveto?
In that case you are right

wfsim/strax_interface.py
petergaemers: WFsim will break instructions up in those which are close together. These groups are simulated until it's over threshold. It wiill include the rest of the group

wfsim/strax_interface.py
petergaemers: total number

wfsim/strax_interface.py
petergaemers: Correct. For nveto standalone this is not needed, you'd simply run RawRecordsFromFaxnVeto directly and you're good

wfsim/strax_interface.py
petergaemers: I'm pretty sure it's not even needed anymore. It's from back in the day when the mailbox timed out after 1 minute or so and assumed everything crashed

wfsim/strax_interface.py
```diff
@@ -100,31 +109,39 @@ def read_optical(c):
...
+    return ins[mask], channels[mask], timings[mask]
+        
+def _read_optical_nveto(config, events):
+    '''Helper function for nveto to read photon channels and timings from G4 and apply QE's
+    :params c: dict, wfsim configuration
+    :params e: g4 root file
+    
+    returns 2 nested arrays
+    '''
     nV_pmt_id_offset = 2000
```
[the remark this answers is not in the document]
petergaemers: At the moment I'm not passing the entire config around to this guy

wfsim/strax_interface.py
```diff
@@ -160,6 +177,168 @@ def instruction_from_csv(filename):
...
+         For the moment we'll just process the whole file and then throw out all events we won't use
+         Epix needs to be imported in here to avoid circle imports'''
+        logging.info("Getting instructions from epix")
+        import epix
+
+        #make deepcopy cause some of the things in the setup will add unhashable stuff
+        epix_config = get_resource(self.context.config['epix_config'],fmt='json')
+        epix_config.update({'input_file':self.context.config['fax_file'],
+                            'entry_start':self.context.config['event_start'],
+                            'entry_stop':self.context.config['event_stop'],
```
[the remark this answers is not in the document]
petergaemers: entry by g4id?

wfsim/strax_interface.py
petergaemers: It already is in wfsim, it's just not used...

wfsim/strax_interface.py
petergaemers: Whoops, good one

zhut19 (approved):

zhut19:
So, the nveto and McChainSimulator code do not have test coverage as there is no rand instruction for those. Which we should keep in mind as a future improvement.

tests/test_wfsim.py
```diff
@@ -66,10 +66,10 @@ def test_sim_nT():
         st = strax.Context(
             storage=tempdir,
             config=dict(
-                nchunk=1, event_rate=1, chunk_size=2,
+                nchunk=1, event_rate=10, chunk_size=5,
                 detector='XENONnT',
                 fax_config=('https://raw.githubusercontent.com/XENONnT/WFSim'
-                '/22004e28421044452f42aaf5797be7186e07bbba/files/XENONnT_wfsim_config.json'),
+                '/a412375ad1fc85f30596b2b73cd8ffe5401de42e/files/XENONnT_wfsim_config.json'),
```
jorana, line 72: @petergaemers , I decreased this to 1 and 2 respectively to reduce the overall time we spend on CI. I figured that if it would work for a little data, it would also work for more but with less waiting time, are there many things that will only show up with running more simulations (`event_rate=10, chunk_size=5,`)?

jorana: (stupid github makes it unclear what I was referring to but it's the:
`event_rate=1, chunk_size=2` -> `event_rate=10, chunk_size=5`

for both the nT and 1T test

zhut19: The reason is that I saw some of the S1 code not covered by the test but should be. And it's possible that LCE made S1 too small. So actually be a better solution than increase the event rate is to fix rand instruction. So we can guarantee S1 has detectable photons.

jorana: I see, makes sense. We could also increase the energy, if we can reduce the testing time, there is clear benefit.

petergaemers: I'll make the default energy in rand_instructions larger, s1s are so fast it will not matter

wfsim/core.py
```diff
@@ -491,7 +489,6 @@ class S2(Pulse):
 
     def __init__(self, config):
         super().__init__(config)
-        # TODO
         # This config is not set for the 1T fax config
         self.config.setdefault('s2_time_spread', 1)
```
jorana: If we remove the #TODO, can we also remove these two lines?

wfsim/core.py
```diff
@@ -283,7 +283,6 @@ class S1(Pulse):
 
     def __init__(self, config):
         super().__init__(config)
-        # TODO
         # This config is not set for the 1T fax config
         self.config.setdefault('s1_decay_spread', 1)
         self.phase = 'liquid'  # To distinguish singlet/triplet time delay.
```
jorana: Also for these lines (like the comment below), if we remove the #TODO, can we also remove these two lines?

wfsim/core.py
petergaemers: Both should be removed,. I did some github magic by accident causing them to reappear

jorana: Thanks!

PR #104, opened 2021-02-26
Config patch and debugging prints

jorana:
Due to https://github.com/XENONnT/straxen/pull/384 were infinity hanging on zipped json files. While debugging this I added some debug statements here and there to pin point the issue, I think this generally can benefit others from debugging wfsim too.

In this PR:

- We change the requirements to work on the latest version of straxen with the right commented-json file handling
- We add a file .plintrc that is the configuration for the codefactor tests
- Add some debug statements

- [x] To get this working, we will need a new straxen release first as well as a merge of the strax_aux_files:


jorana:
We need to update the link to the right commit message of the strax-auxfiles

ramirezdiego (approved):

PR #106, opened 2021-03-01
Bug fix for endtime of truth

mzks:
At here
https://github.com/XENONnT/WFSim/blob/3fbae1f9478401fe318b51b27b2b78df0324b833/wfsim/core.py#L1220

endtime of truth has been overwritten to the initial instruction value (0) for all time.
Thus, we always got 0 in truth. 

I would like to check for other parameters not to do like that. (I just checked `endtime`.)

### DIFF
--- a/wfsim/core.py
+++ b/wfsim/core.py
@@ -1215,6 +1215,8 @@ class RawData(object):
                 tb[field] = np.mean(value)
             elif len(instruction) > 1 and field == 'amp':
                 tb[field] = np.sum(value)
+            elif field == 'endtime':
+                pass
             else:
                 # Cannot summarize intelligently: just take the first value
                 tb[field] = value[0]


mzks:
`e_dep` and `vol_id` are also always 0 in nVeto optical simulation, but it doesn't come from the overwritten issue. I've checked the other variables of truth are correct.

mzks:
By the way, the `endtime` is slightly tricky.
This is generated pulses for nVeto.
image

The `time` of truth points the initial time of the initial pulse, but the`endtime` of the truth points the last photon timing. I think they are not consistent. Is it better that `endtime` points the the last time of the last pulse? How do you think about that? If we change (or add as an alternative name) the variable, it is useful when we match the original G4 event to records. Of course, we can match TPC-veto events after that.

petergaemers:
Thank you Keita. So my guess is it's most correct to have endtime be the most right pulse sample. Can you take care of that?

petergaemers:
Actually since the issue comes from the endtime now being in the instructions (this was needed for epix but we dropped that idea) it's maybe nicer to remove this: https://github.com/XENONnT/WFSim/blob/841cef6dbdb23ddf740b4f6a721c0047e5664f96/wfsim/strax_interface.py#L23

petergaemers:
Fix included in https://github.com/XENONnT/WFSim/pull/103

PR #105, opened 2021-02-27
Add a website for documentation on WFSim

jorana:
# Add a website for WFSim documentation.

In this PR:

- I set up a website for WFSim using similar infrastructure to straxen.
- Remove some notebooks that are so out of date that it makes no sense to keep them around.
- Set up a test website to look at the documentation:
  - https://wfsim-test.readthedocs.io/en/latest

## How does this work
Using https://www.sphinx-doc.org and https://readthedocs.org/ a website can be set up in a similar fashion to strax and straxen. I at first did it without the straxen-way but I think it makes sense to abide to a common format so I copied quite some functionality.

## Auto-generated documentation 
[WFSim documentation.pdf](https://github.com/XENONnT/WFSim/files/6054273/document.pdf)

## Further work
- Add a badge to the website
- Remove the other notebooks not linked on the website or add those notebooks
- Mention NV integration
- Change https://wfsim-test.readthedocs.io/en/latest to https://wfsimreadthedocs.io/en/latest

PR #107, opened 2021-03-02
New default nT config file

ramirezdiego:
Simplest PR ever. Default config file for nT sims has now a different name. Explanation here: https://github.com/XENONnT/private_nt_aux_files/pull/24#issuecomment-788732974

`Fax_Tutorial.ipynb` is too outdated and will anyway be renamed in #105, so I didn't touch it.

### DIFF
--- a/notebooks/Understanding_WFSim_P1.ipynb
+++ b/notebooks/Understanding_WFSim_P1.ipynb
@@ -27,7 +27,7 @@
    "outputs": [],
    "source": [
     "config = straxen.get_resource('https://raw.githubusercontent.com/XENONnT/'\n",
-    "                 'strax_auxiliary_files/master/fax_files/fax_config_nt.json', fmt='json')\n",
+    "                 'strax_auxiliary_files/master/fax_files/fax_config_nt_design.json', fmt='json')\n",
     "config.update({'detector':'XENONnT', 'right_raw_extension':50000})"
    ]
   },
@@ -410,7 +410,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.12"
+   "version": "3.8.5"
   }
  },
  "nbformat": 4,
--- a/notebooks/Understanding_WFSim_P2.ipynb
+++ b/notebooks/Understanding_WFSim_P2.ipynb
@@ -28,7 +28,7 @@
    "outputs": [],
    "source": [
     "config = straxen.get_resource('https://raw.githubusercontent.com/XENONnT/'\n",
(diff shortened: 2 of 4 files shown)


jorana (changes requested):
Hi Diego, to my surprise I learned I've been neglecting your PR and had not noticed that you asked me to review this.

However, the link you added doesn't exist:
https://raw.githubusercontent.com/XENONnT/strax_auxiliary_files/master/fax_files/fax_config_nt.json

We can either: link to the private repo, link to the public repo and use 1T, rely on the downloader, or wait for #111. I think for the tutorials we can rely on the downloader wheres for the tests, #111 should suffice. Do you agree?

ramirezdiego:
Yes, agreed. This came before #111 but we can fix all ends in that PR.

However, notice that the file I added is not the one you link, but `fax_config_nt_design`. It worked a few days ago, but it looks like the token changed, so the link in `strax_interface.py` brings you nowhere.

notebooks/Understanding_WFSim_P1.ipynb
```diff
@@ -27,7 +27,7 @@
    "outputs": [],
    "source": [
     "config = straxen.get_resource('https://raw.githubusercontent.com/XENONnT/'\n",
-    "                 'strax_auxiliary_files/master/fax_files/fax_config_nt.json', fmt='json')\n",
+    "                 'strax_auxiliary_files/master/fax_files/fax_config_nt_design.json', fmt='json')\n",
```
jorana, line 30: I got the link from there

jorana (commented):
> 
> 
> Yes, agreed. This came before #111 but we can fix all ends in that PR.

Absolutely, my bad

> However, notice that the file I added is not the one you link, but `fax_config_nt_design`. It worked a few days ago, but it looks like the token changed, so the link in `strax_interface.py` brings you nowhere.

I agree

notebooks/Understanding_WFSim_P1.ipynb
jorana: to be fair, the line from before also doesn't work

ramirezdiego: Ah, oops 😄 I thought you were just looking at the script. Indeed in the notebooks I just changed `nt` by `nt_design` from the command line, so completely overlooked that this link was too old to work (`fax_files` instead of `sim_files`, public repo...).

I close this PR and will fix the tutorials once #111 is in place, since I think other settings need to be updated besids this one.

PS: Shame on me, when you read the first three words I wrote in this PR.

jorana: Sounds good, I think that would make sense, thanks Diego!

Issue #89, opened 2021-02-11
New runID argument to communicate with the CMT


ramirezdiego:
Ideally, the user would provide a DAQ run ID and we'd query the corrections at that stage, updating the alerady existing CMT interface in WFsim. For the CMT/straxen experts: which complications do you foresee here?

jorana:
Actually, this request is better suited in the straxen repo because CMT is not actually a relevant part of WFSim.

>  For the CMT/straxen experts: which complications do you foresee here?
Very little, the most complicated part is most likely having a "political" agreement on:
- Consistent run naming in MC
- What gains to use (what run), when to change this, where to document this
- Actually all using this convention instead of starting a context every time from scratch (i.e. using a common context)

Also tagging @ahiguera-mx because he is the one doing all this CMT stuff.

ramirezdiego:
> Actually, this request is better suited in the straxen repo because CMT is not actually a relevant part of WFSim.

Thanks Joran. You're right, but I thought the run id argument would be implemented only in WFsim (since otherwise it is clear that the CMT should read the processed run id, and not a constant value).

About the agreement:
- I think we can diverge in run namings (i.e., add non-integers to MC runs) when the option to call the CMT is implemented in the proposed way.
- This is the main question indeed, but I don't have an answer as I am not familiar with tha gains fluctuation. @ahiguera-mx?
- This is the intention.

ahiguera-mx:
I think there is a potential solution for this without modifying CMT, as I understand WFsim want to defined a given `run_id` different from DAQ `run_id` this make total sense, however that would prevent the use of CMT to retrieve actual PMT gains from the corrections DB. To solve this I propose the following approach. 



```
cmt=straxen.CorrectionsManagementServices()

st = strax.Context(
        storage=strax.DataDirectory('/home/'),
        config=dict(detector='XENONnT',
                    fax_config='fax_config_nt.json',
                    check_raw_record_overlaps=False,
                    **straxen.contexts.xnt_common_config,
                    ),
        **straxen.contexts.xnt_common_opts)

# This line will get actual PMT gains from a given run_id: gains_run_id is a run_id from RunsDB
gains = cmt.get_pmt_gains(gains_run_id, 'to_pe_model', 'ONLINE')
straxen.FIXED_TO_PE['WFsim_gains'] = gains

st.set_config(dict(gain_model=('to_pe_constant', 'WFsim_gains')))

st.register(wfsim.RawRecordsFromFaxNT)
```

[...]

PR #108, opened 2021-03-02
Adding gain_run_id context cfg

zhut19:
Addressing the issue in https://github.com/XENONnT/WFSim/issues/89

Also a very simple pr.


ramirezdiego:
Thanks for pushing this!! Since the fix is quite trivial indeed and is known by the WFSim (current) users, I wonder if we can already implement the general `run_number` query from the `fax_config*` file, as I mentioned in Slack.

The idea is that first one needs to load the context and point to a `fax_config*`. If the `run_number` (maybe a better name for this is needed) of the config file equals `0`, then defaults from the file itself and WFSim are loaded for gains, lifetime, etc. If `run_number=!0`, then we fetch these parameters from the CMT with a query analogous to that for the gains.

Does this make sense?

We would just make sure to add a "blessed" run number in the config files and specify what this parameter stands for in the comments.

zhut19:
@ramirezdiego , I still think 'cmt_run_id' is more clear than just 'run_id' and since it's a string, I like using 'id' instead of 'number' just as a habit from pax time. 
I leave the get_elife as a todo, as I don't think it's finalized in straxen yet.

ramirezdiego:
> 
> 
> @ramirezdiego , I still think 'cmt_run_id' is more clear than just 'run_id' and since it's a string, I like using 'id' instead of 'number' just as a habit from pax time.
> I leave the get_elife as a todo, as I don't think it's finalized in straxen yet.

Agreed! I was calling it `run_number` before because that's the name in the config file, but let's change it to the one you implemented now.

wfsim/strax_interface.py
```diff
@@ -320,8 +320,12 @@ def __init__(self, config):
     strax.Option('fax_config',
                  default='https://raw.githubusercontent.com/XENONnT/private_nt_aux_files/master/sim_files/fax_config_nt.json?token=AHCU5AZMPZABYSGVRLDACR3ABAZUA'),
     strax.Option('gain_model',
-                 default=('to_pe_per_run', 'https://github.com/XENONnT/private_nt_aux_files/blob/master/sim_files/to_pe_nt.npy?raw=true'),
+                 default=('to_pe_per_run',
+                          'https://github.com/XENONnT/private_nt_aux_files/blob/master/sim_files/to_pe_nt.npy?raw=true'),
```
jorana, line 324: This doesn't actually work right?

zhut19: Right, this is not working. Any suggestions?, I am hoping to fix this some other time when we want to bring the default setting back to work again.

jorana: I think `('to_pe_constant', 0.01)` like https://github.com/XENONnT/straxen/pull/363/files is most transparent especially if all is overwritten with some CMT value anyway

wfsim/strax_interface.py
```diff
@@ -353,19 +357,20 @@ class FaxSimulatorPlugin(strax.Plugin):
...
-        # Update gains to the nT defaults
-        self.to_pe = get_to_pe(self.run_id, c['gain_model'],
+        overrides = self.config['fax_config_override']
+        if overrides is not None:
+            c.update(overrides)
+
+        cmt_run_id = c['cmt_run_id'] if c['cmt_run_id'] != '0' else self.run_id
+        # Update gains depending on gain model
+        self.to_pe = get_to_pe(cmt_run_id, c['gain_model'],
                               c['channel_map']['tpc'][1]+1)
```
ramirezdiego, line 367: If I understand correctly how the `get_to_pe` function works, I think there are two potential errors in here that we want to avoid:
1. If `cmt_run_id!=0` and `self.run_id` is not an integer (and it probably won't), the run will crash because CMT expects integers.
2. If `cmt_run_id==0` and `self.run_id` is an integer, the simulation file has probably the wrong name, and the user mostly likely doesn't want to query the gains from run `self.run_id`.

Can we just change this logic to something like ... ?
 ```
cmt_run_id = c['cmt_run_id']
if cmt_run_id!=0:
   self.to_pe = get_to_pe(cmt_run_id, c['gain_model'],...)
 ```
 
 and otherwise nothing is done, i.e. the default `gain_model` loaded with the context remains.

zhut19: I think it is fine, `get_to_pe` and `get_corrections_config` in general takes run id like `'012605'`. I guess I can add an assertion just to be safe.
`assert isinstance(cmt_run_id, str), "cmt_run_id should be a string not %s" % type(cmt_run_id).__name__`

wfsim/strax_interface.py
zhut19: Okay

wfsim/strax_interface.py
ramirezdiego: I am not sure this would solve the issue. Let me drop some examples of the cases 1 and 2 I wrote above:
1. `cmt_run_id!=0` & `self.run_id=mc_run_0001` -> CMT crashes because this is a string, but needs to be converted into an int at some point.
2. `cmt_run_id!=0` & many int MC run_ids (many jobs running at the same time for a single simulation!): `self.run_id=00001`, `self.run_id=00002`, ..., `self.run_id=10000` -> In this case we don't want to call the CMT for all these different runs, so `self.run_id` must play zero role in the CMT call.

That's why I think that the only option is to have always some working defaults (as you just set with Joran), and just call CMT for `cmt_run_id!=0`.

Does this make sense?

ramirezdiego:
Hey @zhut19, then we go for something like:
```
if cmt_run_id=!0:
   c['gain_model']='CMT_model'

self.to_pe = get_to_pe(cmt_run_id, c['gain_model'])
```
where `cmt_run_id` has only a use in the CMT model case?

WenzDaniel:
Hej, I just had a longer discussion with Diego about this issue. I think there wont be anyway to solve this, but to change CMT slightly. The issue is the following `self.run_id` will always be associated with the generalized run_id string. Let us say for the sake of an example we create `st.get_array('mc_002',  'event_info')` then `self.run_id = 'mc_002'` which means that we pass in every [straxen.get_to_pe](https://github.com/XENONnT/straxen/blob/e0bf0f21c797ec0008bc13b779d6c00a6ae4fcbb/straxen/plugins/peaklet_processing.py#L106-L108) call a string as run_id. Hence I would propose to introduce a new gain_model supported by CMT which would look like: 
```
gain_model = ('WFSim', ('015600', 'to_pe_model', 'online'))
elife_model = ('WFSim', ('015300', 'elife_model', 'online'))
```
And in [get_to_pe](https://github.com/XENONnT/straxen/blob/e0bf0f21c797ec0008bc13b779d6c00a6ae4fcbb/straxen/get_corrections.py#L11) we would have to replace:
```
model_type, model_conf = gain_model
```
with:
```
if gain_model[0] == 'WFSim':
    run_id = gain_model[1][0]
    model_type = CMT_model
    model_conf = gain_model[1][1:]
else:
    model_type, model_conf = gain_model
```
Maybe an even better approach would be to write a decorator which translate every correction input from a WFSim notation to the normal notation. Maybe this is an opportune moment to generalize this issue and reorganize the way we call our corrections. In the beginning it was not fully clear which different cases we would need, but now we have a much better overview. E.g. I do not see why we would need both `CMT_model` and `to_pe_model`. Let me just tag some relevant people who might want to discuss. @ahiguera-mx @tunnell @jorana @zhut19 @ramirezdiego

ahiguera-mx:
> Hej, I just had a longer discussion with Diego about this issue. I think there wont be anyway to solve this, but to change CMT slightly. The issue is the following `self.run_id` will always be associated with the generalized run_id string. Let us say for the sake of an example we create `st.get_array('mc_002', 'event_info')` then `self.run_id = 'mc_002'` which means that we pass in every [straxen.get_to_pe](https://github.com/XENONnT/straxen/blob/e0bf0f21c797ec0008bc13b779d6c00a6ae4fcbb/straxen/plugins/peaklet_processing.py#L106-L108) call a string as run_id. Hence I would propose to introduce a new gain_model supported by CMT which would look like:
> 
> ```
> gain_model = ('WFSim', ('015600', 'to_pe_model', 'online'))
> elife_model = ('WFSim', ('015300', 'elife_model', 'online'))
> ```
> 
> And in [get_to_pe](https://github.com/XENONnT/straxen/blob/e0bf0f21c797ec0008bc13b779d6c00a6ae4fcbb/straxen/get_corrections.py#L11) we would have to replace:
> 
> ```
> model_type, model_conf = gain_model
> ```
> 
> with:
> 
> ```
> if gain_model[0] == 'WFSim':
>     run_id = gain_model[1][0]
>     model_type = CMT_model
>     model_conf = gain_model[1][1:]
> else:
>     model_type, model_conf = gain_model
> ```
> 
> Maybe an even better approach would be to write a decorator which translate every correction input from a WFSim notation to the normal notation. Maybe this is an opportune moment to generalize this issue and reorganize the way we call our corrections. In the beginning it was not fully clear which different cases we would need, but now we have a much better overview. E.g. I do not see why we would need both `CMT_model` and `to_pe_model`. Let me just tag some relevant people who might want to discuss. @ahiguera-mx @tunnell @jorana @zhut19 @ramirezdiego

WenzDaniel:
> Regarding the change, I would rather go for the option cmt_run_id instead of changing CMT
As I said this wont work without changing either the base plugin class in strax or every setup in every straxen plugin.

ahiguera-mx:
> > Regarding the change, I would rather go for the option cmt_run_id instead of changing CMT
> 
> As I said this wont work without changing either the base plugin class in strax or every setup in every straxen plugin.

But that is because the way WFSim defines "run_id", right not because tools or plugins. Or am I missing something fundamental here

ramirezdiego:
@ahiguera-mx WFSim doesn't define any run_id, this is set via strax context.

ahiguera-mx:
> @ahiguera-mx WFSim doesn't define any run_id, this is set via strax context.

Then what do you guys mean by `'mc_002`?

ramirezdiego:
We want to give a name to our output files (G4+WFSim), and we chose `mc_002` to be that name. Then strax knows that the output should be called, e.g., `mc_002-raw_records`. But we don't want `mc_002` to be what communicates with CMT. Hence the `cmt_run_id` solution, that can only work under the conditions described by Daniel.

WenzDaniel:
`mc_002` is the run_id specified for the particular simulation. So it is an "identifier" similar to our run_ids for the TPC. In the end a single simulation e.g. neutron calibration will be made up by many of these runs e.g. `mc_001` till `mc_100` and each of these simulations should use the correction from a specific a real run e.g. 015600. But in straxen you will load the simulated data as `st.get_array('mc_002', 'raw_records')`.

ahiguera-mx:
That I understand, What I'm saying is that WFSim can configure the strax options that are store in CMT independently of the identifier, right? and this hast to be done only once during processing, right?

WenzDaniel:
> That I understand, What I'm saying is that WFSim can configure the strax options that are store in CMT independently of the identifier, right? and this has to be done only once, during processing, right?

No, we call CMT also in higher plugins e.g. in peaklet as I linked above. Hence once you created raw_records with the run_id `mc_002` it will keep this string as a run_id also for higher plugins and as such we run into an issue when executing: https://github.com/XENONnT/straxen/blob/e0bf0f21c797ec0008bc13b779d6c00a6ae4fcbb/straxen/plugins/peaklet_processing.py#L106-L109 in every plugin which loads gains.

ahiguera-mx:
I thought peaklets is part of the processing... Anyway, let's discussed it tomorrow. Maybe a better solution is make CMT part of the context so it can be configured all at once

zhut19:
> That I understand, What I'm saying is that WFSim can configure the strax options that are store in CMT independently of the identifier, right? and this has to be done only once, during processing, right?

Maybe this is not clear, WFSim is used the same way as those processing plugins that are registered in a strax context. Not the other way around, creating strax context inside WFSim. You can find a simple example [here](https://github.com/XENONnT/WFSim/blob/master/tests/test_wfsim.py#L66-L82).

ramirezdiego:
Resolved in https://github.com/XENONnT/straxen/pull/443, following the discussion above and the one here: https://xenonnt.slack.com/archives/C017EMG8QAK/p1618579531066100

PR #109, opened 2021-03-05
allow ntauxfiles as a backup option for developers

jorana:
Allow developers to be on a specific ntauxfiles branch and load resources from there to develop wfsim

**WFSim needs you:**
 - [x] Please add a test for this PR, as a bare minimum, make sure it's covered in coveralls!
 - [x] Can you add a docsting to all your functions?


petergaemers (approved):
Thanks Joran :)

PR #110, opened 2021-03-09
Base configs

zhut19:
Adding WFSim/files folder for some fake configs for nT. Thus the testing and default resource fetching can avoid requesting from the private repository.


jorana:
Many thanks Tianyu, I highly support this approach because I think it is nice to have some "untuned" but working example of how physics is modeled by WFSim even if the parameters are dummy parameters.

Please don't hate me for saying so, but in principle placeholder files could be added to our public file repo: https://github.com/XENONnT/strax_auxiliary_files/. Here are some pros/cons I quickly came up with:

Pros:
- WFSim stays lean, you might otherwise end up like infinite waiting on `git pull` and the whole package becoming clunky.
- WFSim keeps it's focus on code, not files.
- It is more consistent with how 1T files are handled.

Cons:
- You are every time working somewhere else when updating some tweak of WFSim which is slightly annoying.
- This is the `x`th suggestion by a non-dev while already `y` contradictions suggestions were made, let's just get it over with.
- In case you want to auto-doc something based on the config, you might want to have these files in the repo (perhaps?) such that it's easier to explain what does what etc. Fetching it from a remote place is not super hard on the other hand.
- If you want something from not-quite master, you have to catch the annoying git-hash and patch that in the url, this is annoying.
- Versioning is harder when slight tweaks might be referring to another thing when trying to load later from master.

Break even:
- Both repos are public, rendering any discussion in this direction moot.

Perhaps you have some other motivations not listed here. Based on the above I could imagine that moving the Maps etc to the aux-file repo makes sense, those change probably never. The config.json is a different story, I could argue both ways as you see from my pros/cons. Hope this helps.

zhut19:
> Reminder to not merge this PR before #103, since it is built upon that branch.

Actually, I am thinking to merge this first and merge master to https://github.com/XENONnT/WFSim/pull/103 (I can resolve the conflicts during the latter). Otherwise, I don't see how the S2 refactor can pass the pytest.

ramirezdiego:
> > Reminder to not merge this PR before #103, since it is built upon that branch.
> 
> Actually, I am thinking to merge this first and merge master to #103 (I can resolve the conflicts during the latter). Otherwise, I don't see how the S2 refactor can pass the pytest.

This was also my thought of the plan we discussed on Friday, but then I don't understand why you built the changes on top of Peter's branch. Maybe just a mistake? If we merge this PR, we'll be merging both, #103 and this one. Probably the best is to close this PR and to submit your changes on top of master.

petergaemers:
Since this guy includes most of #103, I've split this guy to only include the config changes here #111

PR #111, opened 2021-03-11
Base configs reloaded

petergaemers:
PRs 110 and 103 are mixed, here I've tried to seperate them


zhut19:
Adding to the discussion posted https://github.com/XENONnT/WFSim/pull/110, we can avoid the S1 and S2 pattern maps during testing, but just using the DummyMap class to return a constant value. This benefits us because those two maps are relatively large.
But I still think it makes life easier to put a config here that we don't need to do pull requests in two repos at the same time.
While putting the SPE here is me being lazy.

wfsim/core.py
```diff
@@ -407,8 +417,7 @@ class S2(Pulse):
 
     def __init__(self, config):
         super().__init__(config)
-        # TODO
-        #  This config is not set for the 1T fax config
+        # This config is not set for the 1T fax config
         self.config.setdefault('s2_time_spread', 1)
```
jorana, line 421: Perhaps this is a good moment to remove this line?

wfsim/load_resource.py
```diff
@@ -113,49 +122,96 @@ def __init__(self, config=None):
             self.s2_light_yield_map = make_map(files['s2_light_yield_map'], fmt='json')
             self.s2_pattern_map = make_map(files['s2_pattern_map'], fmt='json.gz')
             self.fdc_3d = make_map(files['fdc_3d'], fmt='json.gz')
-            # TODO
-            #  config not set
-            self.gas_gap_length = lambda positions: np.ones(253)
+
+            # Gas gap warping map
+            if config['enable_gas_gap_warping']:
+                self.gas_gap_length = make_map(["constant dummy", 1, [254,]])
```
jorana, line 128: nice, makes more sense

jorana (approved):
Hi Tianyu just scanning the code it looks good, I think it is fine to indeed have a base config in WFSim

wfsim/core.py
```diff
@@ -387,10 +387,20 @@ def er(size,config, singlet_triplet_delays, phase):
...
+
+        timings[primary] += Pulse.singlet_triplet_delays(
+            size_primary, config['s1_ER_primary_singlet_fraction'], config, phase)
+
+        # Correct for the recombination time
+        # For the non-exponential distribution: see Kubota 1979, solve eqn 2 for n/n0.
+        # Alternatively, see Nest V098 source code G4S1Light.cc line 948
+        timings[~primary] *= 1 / (-1 + 1 / np.random.uniform(0, 1, size - size_primary))
+        # Update max recombine time in the nT fax config
+        config['maximum_recombination_time'] = 1000
```
jorana, line 399: perhaps do `config.setdefault('maximum_recombination_time', 1000)`, Is this the right place by the way?

zhut19: Unfortunately, it has been set in all our config files, but it seems that the original value is a bit too tight. We can remove this line but remember to change in the other files.

PR #116, opened 2021-03-16
Reintroduce self._electron_timings for truth

petergaemers:
The truth tries to readout a variable self._electron_timings for the number of produced electrons. This PR reintroduces this variable since it wasn't stored anymore in pr #103


zhut19 (approved):
Can I merge this?

PR #117, opened 2021-03-16
Fix PMT Afterpulse

zhut19:
Debugging of PMT afterpulse class


petergaemers:
Hi Tianyu, is this finished or still a work in progress?

PR #118, opened 2021-03-16
maybe fixing chunking issue

zhut19:
Like the name of this pr


petergaemers:
I was looking at the same place :P #119

terliuk:
i will try to have a look today. It is a bit tricky to be sure that it doesn't happen again, as there is about 3-5 jobs out of 100 that fail due to this.

zhut19:
@terliuk now I don't think this pr is ready, one of the tests failed due to that a chunk is created before the previous one end :(.

zhut19:
@terliuk are you still available for a quick review? I tried with ~100 random seeds, those two changes are the only issue I catch.

PR #119, opened 2021-03-17
Fix the occasional 'data ouside chunk boundary'

petergaemers:
This PR fixes the very occasional "data outside chunk boundary by 10ns"


petergaemers:
Same as #118

PR #121, opened 2021-03-24
Fix non-literal compare for str

jorana:
```python
/home/angevaare/software/WFSim/wfsim/strax_interface.py:645: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  for data_type in self.provides if data_type is not 'truth'}
```
Fix this warning


petergaemers:
Thanks Joran

petergaemers (approved):

PR #123, opened 2021-03-24
update history of releases

jorana:
For e.g. https://pypi.org/project/wfsim/ it is nice to keep a record of the releases published. This is based on the history. This PR updates that history for the last two releases.


ramirezdiego (approved):
Let's merge this. I think everyone will agree.

PR #124, opened 2021-03-25
Pass full config to Resource for read_optical_nveto

petergaemers:
Load_resource is called by the nveto loader to get the pmt qe's. Because historical reasons we only passed a very small config around. This is no longer needed


ramirezdiego (changes requested):
Hey Peter, I know you want to keep this PR separated from https://github.com/XENONnT/WFSim/pull/125 for a cleaner commit log (which makes sense), but before merging this fix I would make sure that this line, in addition to the changes in https://github.com/XENONnT/WFSim/pull/125, makes the full_chain WFSim class work. As you know, you would need to then run the `run_wfsim.py` script from the mc_chain branch you created: https://github.com/XENONnT/mc_chain/pull/12. I know this is a bit decoupled from the mc_chain life, but just in case other related nVeto issues arise, as I suspect.

Let's chat tomorrow and I'll share the tests I ran.

petergaemers:
Ah scratch this. It's still needed

PR #127, opened 2021-03-26
Load GARFIELD map according to liquid level

ramirezdiego:

I pushed the changes in the `Resource` class for loading the GARFIELD maps according to the simulated liquid level, respecting the format introduced in https://github.com/XENONnT/private_nt_aux_files/pull/46. I found this was a better place than the ` luminescence_timings_garfield` function, but let me know if you disagree.

I also changed the default map to load.


ramirezdiego:
(In case it wasn't clear: to merge only after https://github.com/XENONnT/private_nt_aux_files/pull/46).

zhut19 (approved):

PR #130, opened 2021-03-28
Rz maps

zhut19:
 - Implement field-dependent radial-depth maps including `survival_probability_map`, `drift_speed_map`, `diffusion_longitudinal_map`
 - Two small fixes for scintillation timing, its dtype and moving the single electron center time to zero.


wfsim/core.py
```diff
@@ -608,9 +646,10 @@ def _luminescence_timings_simple(n, dG, E0, r, dr, rr, alpha, uE, p, n_electron,
             ne = n_electron[i]
             dt = dr / (alpha * E0[i] * rr)
```
ftoschi, line 647: I am trying to follow the code here: E0 should be the constant electric field where the anode contribution is not dominating, dr the z step and rr the inverse of the position (constant if outside the anode-dominated field). If in the constant field region, isn't rr changing the value of the field, as it would be clipped to 1 / rA?

wfsim/core.py
```diff
@@ -608,9 +646,10 @@ def _luminescence_timings_simple(n, dG, E0, r, dr, rr, alpha, uE, p, n_electron,
...
             dy = E0[i] * rr / uE - 0.8 * p  # arXiv:physics/0702142
+            avgt = np.sum(np.cumsum(dt) * dy) / np.sum(dy)
 
             j = np.argmax(r <= dG[i])
-            t = np.cumsum(dt[j:])
+            t = np.cumsum(dt[j:]) - avgt
```
ftoschi, line 652: Why do remove the yield-weighted average time? This yields to negative time, is this supposed to happen? Also this average accounts for values outside the physical path of the electron in case of warping (as you take the larges gas gap for r)

wfsim/core.py
zhut19: I don't remember doing this, but during the copy from fax to here, I merged a factor of rA into E0. So now this E0 equals to E (constant between rL and rA) * rA. And thus E0 * rr at inverse rA is that constant.

wfsim/core.py
zhut19: Yeah, we do need this time shift, unfortunately, this is because the gate time we measure from data (the time shift applied to all S2s) is using the S2 center time. If we don't shift SE by its center time here, all the S2 center time will be reproduced a bit late. We will see Z_reconstructed deeper than Z_instruction.

wfsim/core.py
ftoschi: Ah ok, then it makes sense

petergaemers (approved):
Hi @zhut19 is this finished or is there still something missing?

zhut19:
> Hi @zhut19 is this finished or is there still something missing?

No, was just waiting for a bit of approval.

PR #131, opened 2021-04-01
Don't track the timeout

jorana:
The lineage should not depend on this, right?


petergaemers:
Hey Joran. Thanks for the reminder.

This guy is from back in the day that the mailbox timed out in 5 seconds or so. Currently it doesn't do anything. So it's save to remove completly

jorana:
Sounds good, this will change the lineage of course and not very urgent. I think you guys know best when is the opportune moment to merge.

PR #138, opened 2021-04-10
change dtype of truth

jorana:
Change / update the dtype of the truth info.

While working on some documentation (https://pema.readthedocs.io/en/latest/reference/datastructure.html), I noticed that we may add some description to the truth dtype. Additionally, we might save all the times as integers, right?

Please feel free to edit this PR, I could have filed an issue instead to ask the maintainers to do this but I figured I might just as well get started with it.

**edit**
Apparently, `np.nan` is used a lot for time info not being present. I see that this is somewhat of an issue but perhaps it is worth to consider changing the `np.nan` times to some `wfsim` defined `INTNAN = -int(1e9)` or so to signify that it is not a "real" time. I see this getting tricky since nobody is preventing you from simulating data with negative timestamps (i.e. extensions in strax may do this even if you start at `t=0`.

For now, I just reverted back to setting everything back to a `float64` as it were.


zhut19:
This is nice! Thanks @jorana

zhut19 (approved):

PR #141, opened 2021-04-22
Make v0.5.0 release

ramirezdiego:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,21 @@
+v0.5.0
+======
+* McChainSimulator (major) patch (#125)
+* Update utilix version in requirements.txt (ffc6c2f)
+* Don not track the timeout (#131)
+* Change dtype of truth (#138)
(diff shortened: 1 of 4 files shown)


jorana (approved):

HISTORY.md
```diff
@@ -1,3 +1,21 @@
+v0.5.0
```
jorana, line 1: @ramirezdiego , worth adding the data to the title of the release like in https://github.com/XENONnT/straxen/blob/master/HISTORY.md?

jorana (approved):

HISTORY.md
ramirezdiego: Ok, will do. Was respecting the old format, but I guess this helps for PyPi visualization.

jorana: Many thanks, I did the same previously but only realized that I omitted it later

petergaemers:
Thanks Diego :)

PR #143, opened 2021-04-28
Map update followup

zhut19:
After merging https://github.com/XENONnT/private_nt_aux_files/pull/50, in which we split the S2 sum-PMT LCE map out into an independent file and normalize to 1, so that we can keep the s2_secondary_sc_gain as the data-driven value in the unit of  PE (observed) per electron.

We also decide that s2_secondary_sc_gain value contains the factor '1+p_double_pe_emision', thus when calculating the S2 gain we always divide it by '1+p_double_pe_emision'


petergaemers (approved):

wfsim/load_resource.py
```diff
@@ -268,3 +263,29 @@ def reduce_last_dim(self):
         shape[-1] = 1
 
         return DummyMap(const, shape)
+
+
+def clear_cached_configs():
```
ramirezdiego, line 268: @zhut19 Is this function used?

zhut19: Ah, this one is somehow in my local master branch, keep mixing into my or. It's not used anywhere.

ramirezdiego:
@petergaemers You tested with a local copy of the private files, right? I think the new normalized map wasn't uploaded to the database yet, but maybe I am doing something wrong in my context.

wfsim/load_resource.py
ramirezdiego: Ok, will you delete it or want to keep it there for an upcoming cache cleanup PR?

zhut19: Actually, I don't see if this will be used officially in the future. Let me delete it.

ramirezdiego (approved):
Looks good, I simulated some SEs and the size makes sense now :)

PS: The new map still needs to make it to the database, unless I did something wrong with my environment (see comment above to Peter).

PPS: I think it would make sense to make a new patch release, 0.5.1, based only on this change. Just because it resolves the "bug" on the S2 area with default configurations.

petergaemers:
I've updated the database

PR #146, opened 2021-04-30
add tests for #144

jorana:
Add a failing test to not run into #144 again

This should be fixed in #145

### DIFF
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -25,12 +25,12 @@ jobs:
         python-version: 3.8
     - name: Checkout repo
       uses: actions/checkout@v2
-    - name: Install python dependencies
-      uses: py-actions/py-dependency-install@v2
     - name: Install requirements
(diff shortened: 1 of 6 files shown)


jorana:
@petergaemers , you already fixed this in #146 but I still think it's nice to add some tests for it, for example. the nT and 1T context don't work for all their registered plugins apparently.

Maybe a question mostly for my understanding, would it make sense to put the context of the simulations in WFSim rather than straxen? That would solving the tests that are now failing in this PR much easier?

petergaemers:
Thanks joran. I think it makes much more sense to keep all contexts in the same place

petergaemers (approved):

jorana:
> 
> 
> Thanks joran. I think it makes much more sense to keep all contexts in the same place

Alright, I don't have a strong opinion so it's fine for me. Without https://github.com/XENONnT/straxen/pull/461 this test will keep failing though.

PR #125, opened 2021-03-25
McChainSimulator (major) patch

ramirezdiego:
**What is the problem / what does the code in this PR do**
Hunting some bugs in the McChainSimulator class, to facilitate our life on the epix and mc_chain side.

For the moment I just blindly pushed a fix discussed with @petergaemers, since we get `AttributeError: 'McChainSimulator' object has no attribute 'instructions_nveto'` when running the code with nVeto set to `st.set_config(dict(neutron_veto=False))`.

(To be continued)

PR #147, opened 2021-04-30
Fix nveto readout

petergaemers:
Missed old way to refer to writing out nv data.
nv is always in data type whether you want it or not. Use the targets option instead

### DIFF
--- a/wfsim/strax_interface.py
+++ b/wfsim/strax_interface.py
@@ -721,7 +721,7 @@ class RawRecordsFromMcChain(SimulatorPlugin):
                     data=result_nv[data_type.strip('_nv')],
                     data_type=data_type)
                 #If nv is not one of the targets jus return an empty chunk
-                if 'nv' in data_type:
+            if 'nv' in data_type:
                     chunk[data_type] = self.chunk(
                     start=self.sim.chunk_time_pre,
                     end=self.sim.chunk_time,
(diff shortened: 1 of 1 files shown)


ramirezdiego:
Wait, was this supposed to trigger an error? I'm quite sure I tested #125 right before merging. Have you tested now a dummy TPC+nVeto run?

petergaemers:
It crashes for just TPC

petergaemers:
@ramirezdiego Oke this is the final one. Sorry for the confusion

wfsim/strax_interface.py
```diff
@@ -714,13 +714,20 @@ def compute(self):
...
                     data_type=data_type)
-            else:
+                #If nv is not one of the targets just return an empty chunk
+            if 'nv' in data_type:
+                chunk[data_type] = self.chunk(
+                    start=self.sim.chunk_time_pre,
+                    end=self.sim.chunk_time,
+                    data=np.array([]),
+                    data_type=data_type)
+            elif 'nv' not in data_type:
```
ramirezdiego, line 730: I don't understand the logic here. Why not...?
```
if 'nv' in data_type:
    if 'nveto' in self.config['targets']:
        bla
    #If nv is not one of the targets just return an empty chunk
    else:
        blabla
else:
    blablabla
```

The current code would redefine `chunk[data_type]` twice, if nveto is one of the targets, wouldn't it?

ramirezdiego: Fixed here: https://github.com/XENONnT/WFSim/pull/147/commits/322807f2faaeea961d0c40e81454abc846794187

zhut19 (approved):
I tested running with target set to both `('tpc')` and `('tpc', 'nveto')`, both works

PR #148, opened 2021-04-30
S1 timing splines implementation for top/bottom arrays

terliuk:
This code implements splines for timing distribution of S1 photon hits for each array, as was discussed in recent MC telecon.

**WFSim needs you:**

I added comments describing the logic inside the function


petergaemers:
Hi Andrii, thanks a lot. I have a question. So where is the scintillation delay? It looks like it's already included in the spline?

terliuk:
Yes, currently it is included in the spline. In my plans is also to provide propagation time + NEST implementation, but it might need a bit more time to fully implement. Current spline is Geant4 propagation + 32.1 keV Kr83m NEST.

wfsim/core.py
```diff
@@ -373,7 +373,7 @@ def photon_channels(positions, n_photons, config, s1_pattern_map):
         return _photon_channels
 
     @staticmethod
-    def photon_timings(t, n_photons, recoil_type, config, phase):
+    def photon_timings(t, n_photons, recoil_type, config, phase, channels=None,positions=None,resource=None):
```
ramirezdiego, line 376: (Small comment) Since we're here, would be nice to extend a bit on the args description in the docstring below, plus of course add the new args you coded.

terliuk: Done

petergaemers (approved):

Issue #139, opened 2021-04-13
General comments


tunnell:
If you putting it on PyPI, I'd keep the HISTORY file up to date because otherwise the PyPI page when you are looking for the package only shows really outdated irrelevant info.

tunnell:
Also, see this in setup.py:  'Development Status :: 2 - Pre-Alpha'

tunnell:
Also, this all started with trying to figure out why WFsim versions were appended by a "0." for a student.

jorana:
>, I'd keep the HISTORY 
@tunnell , I agree and made a similar comment/PR:
https://github.com/XENONnT/WFSim/pull/123

ramirezdiego:
Agreed as well. No need for an immediate HISTORY update, but a proper one following #125 (and new release) is planned.

jorana:
With the recent releases and #150 this issue seems solved

PR #150, opened 2021-05-03
Update classifiers

jorana:
Update the classifiers in setup.py, see #139 

This PR updates the setup classifiers. As such, this PR is not about code but about documentation for Pipy: https://pypi.org/project/wfsim/


setup.py
```diff
@@ -37,10 +37,13 @@
     packages=['wfsim',
               'wfsim.pax_datastructure',],
     classifiers=[
-        'Development Status :: 2 - Pre-Alpha',
+        'Development Status :: 4 - Beta',
```
jorana, line 40: another status may also be applicable so please correct me if I am wrong:
https://pypi.org/classifiers/

petergaemers (approved):

PR #152, opened 2021-05-04
Add tests

petergaemers:
This PR adds tests. Since they require acces to the database/private_nt_aux_files you can only run them from somewhere where you do have them. But it will make life easier because it's a smaller manual step than to do it from scratch all the time.

To run them do "pytest" in the wfsim/test folder on dali

Codefactor missed the second commit for some reason


jorana:
@petergaemers this is the direction we love to see:
afbeelding


We also clearly see why numba is our friend. Coveralls will run with numba disabled, thereby increasing the testing duration by a factor of 4.

jorana (approved):
Thanks Peter!

PR #153, opened 2021-05-05
Update Getting_started_wfsim.ipynb

GiovanniVolta:
Missing an s in wfsim.strax_interface.rand_instruction
https://github.com/XENONnT/WFSim/blob/a2707a8ac97bbe7b7f2f6cae979971743987318b/wfsim/strax_interface.py#L57

**What is the problem / what does the code in this PR do**

**WFSim needs you:**
 - Please add a test for this PR, as a bare minimum, make sure it's covered in coveralls!
 - Can you add a docsting to all your functions?

PR #154, opened 2021-05-10
fix notebook, evaluate for docs

jorana:
Despite #153 , the docs are still failing and indeed the notebook does not run out of the box. Let's run it such that sphinx does not have to do this on the build and fail.

It might be worth considering putting this notebook in an `.rst` file as there is not that much code but more info about the several simulation plugins.

### DIFF
--- a/notebooks/Getting_started_wfsim.ipynb
+++ b/notebooks/Getting_started_wfsim.ipynb
@@ -18,8 +18,13 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
+   "execution_count": 1,
+   "metadata": {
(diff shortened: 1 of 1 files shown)


petergaemers:
Mine did not run without installing those two guys. Maybe I should have rerun install. Let me check

petergaemers:
Clean environment fails: 

```/bin/sh: sphinx-build: command not found
make: *** [clean] Error 127
make_docs.sh: line 8: sphinx-apidoc: command not found
rm: source/reference/modules.rst: No such file or directory
/bin/sh: sphinx-build: command not found
make: *** [html] Error 127
```

petergaemers (approved):

PR #155, opened 2021-05-10
use database access to increase coverage of tests

jorana:
# What is the problem / what does the code in this PR do
This PR allows WFSim to pull configs form the database in order to be fully tested. This should increase the coverage and make it easier for the developers to spot potential issues as github actions would work just as well as doing local tests.

For example #152 will be much more useful if this PR is merged.

## Developer commitment required
There is a caveat which the developers should be aware of since this will cause work (less then in the current state but still, they are the ones in charge to fix those things).

In order to have the test also being able to run from a remote fork, the tests need to (stay) drafted as if there is a possibility that one does not have access to the database. We make sure this is the case using the famous line:
```python
   if not straxen.utilix_is_configured():
        return
```

The developers should be consent that this will (almost guaranteed) cause confusion if one makes a PR from their own fork, merge it, and the tests fail. Since on ones fork, there was no database access, on master there is, running a new test, which might fail. The developers need to explain/fix this after the PR merge. This perhaps sounds more complex than it actually is. After all, if one reviews it properly there should be no bugs right :wink:? Also keep in mind that with the addition of this PR, you actually find this out immediately whereas in the current state, you would only notice errors when running locally.

### DIFF
--- a/.github/scripts/create_readonly_utilix_config.sh
+++ b/.github/scripts/create_readonly_utilix_config.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+if [ ! -z "$RUNDB_API_URL" ]
+then
+cat > $HOME/.xenon_config <<EOF
+[basic]
+logging_level=debug
+
+[RunDB]
+rundb_api_url = $RUNDB_API_URL
+rundb_api_user = $RUNDB_API_USER_READONLY
+rundb_api_password = $RUNDB_API_PASSWORD_READONLY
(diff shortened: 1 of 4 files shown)


petergaemers:
Thanks Joran, this not having database acces is a huge pain. I'll add a warning to the relevant tests that it's actually not being run

petergaemers (approved):

jorana:
@petergaemers , that is actually quite a good idea. Why did I never think of that for straxen 🤔

PR #157, opened 2021-05-10
Fix load resource

zhut19:
Fixing the issue that loading map data from the direct path is not working.
Reordering the downloading method, using the installed package before using the mongo downloader.

### DIFF
--- a/wfsim/load_resource.py
+++ b/wfsim/load_resource.py
@@ -5,6 +5,21 @@ import numpy as np
 import strax
 import straxen
 import logging
+
+NT_AUX_INSTALLED = False
+STRAX_AUX_INSTALLED = False
+
+try:
+    import ntauxfiles
+    NT_AUX_INSTALLED = True
+except (ModuleNotFoundError, ImportError):
+    pass
+try:
+    import straxauxfiles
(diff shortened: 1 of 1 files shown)


wfsim/load_resource.py
```diff
@@ -101,48 +101,80 @@ def file_config(config):
...
+            log.info(f"Loading {fname} is successfully from {fpath}")
+            return fpath
+        except (ModuleNotFoundError, ImportError, FileNotFoundError):
+            log.info(f"ntauxfiles is not installed or does not have {fname}")
+
+        try:
+            # You might want to use this, for example if you are a developer
+            log.warning(f"Using the public repo to load {fname} locally")
+            import straxauxfiles  # public package for XENON1T
+            fpath = ntauxfiles.get_abspath(fname)
```
jorana, line 139: Also you can use `fname in straxauxfiles.list_aux_files()`  and `fname in ntauxfiles.list_private_files()` to thereby check if the file is in either and remove all this `try: except` logic.

jorana (changes requested):

petergaemers (approved):

PR #158, opened 2021-05-10
Following up the change in pmt ap file

zhut19:
Following up on https://github.com/XENONnT/private_nt_aux_files/pull/56, bin sizes are added to the pmt afterpulsing file. Here we change the hardcoded bin size to the ones saved in the resource file.

TODO:
~Update this file for XENON1T~


petergaemers (approved):

zhut19:
> Just to make sure, we'll first need to merge this guy before this will pass the test? [XENONnT/strax_auxiliary_files#19](https://github.com/XENONnT/strax_auxiliary_files/pull/19)

Probably not, just needed the commit id

PR #159, opened 2021-05-17
Fix missing last instruction

petergaemers:
As noticed by Daniel and Andrii the last instruction was (not always?) read out. Here I add a final digitization step to make sure we get all data


petergaemers:
The tests are failing because zstd fails to instal....

jorana:
@petergaemers , yes we have seen the same in strax and straxen today.

WenzDaniel:
Nice Peter this was quick. Thanks a lot.

zhut19 (approved):

PR #161, opened 2021-05-18
Fixing SPE reconstruction bias due to truncation

zhut19:
@dantonmartin noticed the -7% bias in SPE reconstruction actually comes from rounding down the float number of the SPE pulse, where we try to mimic the behavior of digitization. We have the same issue here, and can be solved by using `np.around` instead of `np.trunc`. This avoids a more complicated fix of sampling gaussian noise and reorders the code in the digitization function.

The updated gain calculation will remove a 0.14% difference going forth and back between `to_pe` and `gain`

### DIFF
--- a/.github/workflows/coveralls.yml
+++ b/.github/workflows/coveralls.yml
@@ -28,6 +28,7 @@ jobs:
     - name: Install requirements
       run: |
         pip install coveralls
+        pip install zstd~=1.4.9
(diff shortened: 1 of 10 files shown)


petergaemers (approved):

petergaemers:
Hi @zhut19. This makes a lot of sense!

PR #162, opened 2021-05-20
Patch 1T custom context

jorana:
## What is the problem / what does the code in this PR do
In https://github.com/AxFoundation/strax/pull/440 I made the context a bit more strict on the allowed config options. If one says that per-run-defaults are not allowed, we can get great speed improvement.

However, if you use 1T plugins (with per run defaults) this will cause errors:
See e.g. the base env:
https://travis-ci.org/github/XENONnT/base_environment/builds/771769395

## Solution
As such, we need be explicit that it is ok to use per run defaults for 1T simulation.


jorana:
By the way, the evidence that master would now be failing:
https://github.com/XENONnT/WFSim/runs/2632408969?check_suite_focus=true

Issue #166, opened 2021-06-03
Example Kr notebook


zhut19:
The gain calculation bug from #161 and solved in #164, made us realized the importance of having a sanity check notebook. One idea is to simulate Kr events and do simple checks like S1 and S2 area distributions, etc.

Jianyu010:
I placed a sample notebook under ```/project2/lgrandi/jylong/file_delivery/kr83m_wfsim.ipynb```. Can you check whether that's what you want? Thanks.

zhut19:
Resolved in #167

PR #167, opened 2021-06-08
Adding Kr83m notebook

zhut19:

Fixing issue #166. @Jianyu010 I uploaded your notebook here with few small changes.

This notebook should generate Kr83m events and histogram them in S1, S2 space.


Jianyu010 (approved):

zhut19:
Resolves #166

ramirezdiego:
Sorry @zhut19 and @Jianyu010, I was traveling yesterday and made it too late for this change. The notebook looks great!
My single comment is that the code is pretty similar to that of an already existing notebook, right? That would be the `Getting_started_wfsim.ipynb` notebook. I guess this is fine, if you want to have a cleaner version just taking care of plotting Kr83m events, for sanity checks. @zhut19 Will you link it in the PR template, so that reviewers use it for their checks? I guess that in future commits we can also add some more explicative comments of the different steps.

Jianyu010:
@ramirezdiego Hi Diego. Yes this notebook was created when I learned WFSim following the example notebook. The modification should be instead of a regular time gap I sampled randomly.

PR #168, opened 2021-06-11
Oversize chunk handling

zhut19:
The chunk handling in case that we need to chunk early due to too many raw records is problematic. This should fix the issue. But, it's still better if the user can reduce the chunk size.


petergaemers (approved):

ramirezdiego:
@zhut19 Just one basic question: does this alter the default chunking size that you implemented [here](https://github.com/XENONnT/mc_chain/blob/3a336c1ba43be4aa800aac1c5cc512e09706d3f4/run_wfsim.py#L80-L85)?

zhut19:
> @zhut19 Just one basic question: does this alter the default chunking size that you implemented [here](https://github.com/XENONnT/mc_chain/blob/3a336c1ba43be4aa800aac1c5cc512e09706d3f4/run_wfsim.py#L80-L85)? I guess not, but just in case...

The chunking there is rather reasonable. This error happened when I was simulating ~300 S2s > 1e6 PE at the cathode (the widest possible S2s)

PR #169, opened 2021-06-14
Introducing sorting by channel number to preserve timing in S1s

terliuk:
If each channel/array has individual timing distribution, it will be effectively scrambled due to call of `unique` function here: https://github.com/XENONnT/WFSim/blob/s1_ch_sorting/wfsim/core.py#L75 . 

To avoid this unwanted behavior, explicit sorting is performed in S1 class.


zhut19:
Thanks, we do need this.

zhut19 (approved):

petergaemers (approved):

PR #172, opened 2021-06-16
Get gains from CMT

jorana:
Some of the aftermath of https://github.com/XENONnT/straxen/pull/457

Also, @zhut19 , I don't think pinning WFSim permanently is desirable: https://github.com/XENONnT/WFSim/commit/3c925f1a209de18998652c294c6c4b3afefaf6f1, I think we need to change that before making a new WFSim release. Would you agree?

straxen 0.19.0 does not exist yet but hopefully soon


wfsim/strax_interface.py
```diff
@@ -400,8 +400,8 @@ def set_config(self,):
             self.config.update(overrides)
 
         # Update gains to the nT defaults
-        self.to_pe = straxen.get_to_pe(self.run_id, self.config['gain_model'],
-                                       self.config['channel_map']['tpc'][1]+1)
+        self.to_pe = straxen.get_correction_from_cmt(self.run_id,
+                               self.config['gain_model'])
```
jorana, line 404: Actually, this is the only change we really require for WFSim to run with straxen 0.19.0, the rest is just changed conventions for the testing.

zhut19 (approved):

Issue #165, opened 2021-06-03
Adding NEST to S1 scintillation models


zhut19:
We should include [nestpy ](https://github.com/NESTCollaboration/nestpy) in our S1 scintillation model, as also suggested by @terliuk 's study.

ramirezdiego:
Related to this issue, I refer to this message: https://xenonnt.slack.com/archives/C0177G05M1P/p1620812142120200?thread_ts=1620811369.118500&cid=C0177G05M1P
Let's please make clear in the config files which paremeters are used for which model.

terliuk:
I made propagation-only timing splines in addition one containing Kr83m delay (format is more or less the same). I will clean up the code that I use to make them in addition to a file with splines. BTW, we decided to go with top/bottom array timing, right? 

Other parts are rather straightforward:
 - passage of local field, number of photons, excitons and deposit energy to S1 class, as this is the information needed by nestpy 
 -  think about the place, where we want to put netspy object, similar to what we have with resources

ramirezdiego:
Cool!

> 
>     * passage of local field, number of photons, excitons and deposit energy to S1 class, as this is the information needed by nestpy
> 

This one points directly to epix. We discussed this in the MC meeting where the splines propagation was presented and (I think) agreeed on not duplicating to much epix does already, right? Meaning that these parameters would take care of time-only in WFSim and not quanta generation. Just making sure we're on the same step.

terliuk:
> This one points directly to epix. We discussed this in the MC meeting where the splines propagation was presented and (I think) agreeed on not duplicating to much epix does already, right? Meaning that these parameters would take care of time-only in WFSim and not quanta generation. Just making sure we're on the same step.
Yes, it will be direct passage from epix during instruction generation and will only be used for generation of scintillation times to individual photons and not actual quanta generation.

terliuk:
I think we can close this issue as resolved, since NEST part was merged to WFsim, right?

PR #173, opened 2021-06-17
S1 scintillation timing from NEST

terliuk:
Introduces new mode for S1 shape where timing is done from timing spline, and scintillation time is received directly from nestpy which will address issue #165


terliuk:
here is the example of running it : https://github.com/XENONnT/analysiscode/blob/master/s1_pulse_shape/S1_WFsim_pulse_shape_NEST_example.ipynb

skazama:
Hi @terliuk, I cannot comment about epix, but I had a look at your code and it looks fine. Also I managed to run the notebook with your branch and could generate Kr83m S1 signals without errors, so maybe we can merge this? (or do we have to wait for @ramirezdiego's ideas?)

ramirezdiego:
Hi @skazama. The code works fine indeed (thanks for jumping in!), but in a recent chat with @terliuk I suggested that it would be better to push the relevant changes to epix before merging this PR; i.e., to output the required information in the instructions (efield, number of excitons...).

This requires to change the instructions format also in WFSim: either by adding another one, with these extra columns, or adding the columns to the default one, but make epix always print out some dummy number for them. In case this wasn't clear: epix reads the format directly from WFSim: https://github.com/XENONnT/epix/blob/43e7e31a05d0ff52863c2017f367bff7eb859312/epix/io.py#L202 This would be the missing part of this PR.

terliuk:
after recent changes, i need to do extensive testing again, confirming that it works as intended

zhut19 (approved):

PR #175, opened 2021-06-18
Update strax_interface.py

petergaemers:
Update get_to_pe since it's been removed and changed to get_correction_from_cmt


jorana:
https://github.com/XENONnT/WFSim/pull/172 was for the same purpose 😉

PR #177, opened 2021-06-22
Github McChain test using G4 file

mzks:
This PR add test for McChain.
As we discussed before, it is better to have auto-test for MC-Chain WFSim with Geant4 file.
This is also first test for nVeto WFSim.

For this test, this repository includes small geant4 output which has 10 events without useless branches or parameters.
In addition, I added some packages to run in `requirements.txt`.


mzks:
@ramirezdiego and friends,

I made this PR because I would like to avoid crash of nVeto generation in each update. I've tried to run broken nVeto parts many, many times after updates. It makes me tired too much, and it doesn't contributes the progress because it will break again in the next update.
However, current straxen update breaks this PR. Who will fix for that? In this case, that may be me. I would like to know how do people think about that.

mzks:
Thank you very much. Now this PR is ready to merge.

ramirezdiego (approved):
Thanks @mzks. From my side your changes are ready to be merged, and will indeed help for the mc_chain development to have this running automatically.
I leave to @zhut19 or @petergaemers to comment on the format. I see you uploaded a testing ROOT file with no description,n but I assume that's fine for this purpose.

Please remember to tune the McChain test when the new simulations context is released.

Issue #170, opened 2021-06-14
Split and refactor core.py


terliuk:
Our core.py is unreasonably long and very hard to navigate. Does it make sense to split it into separate files corresponding to pulse class, S1, S2 etc? I think it will hugely improve development and testing of our MC.

petergaemers:
Since it's not a lot of work it's fine by me if it helps people

zhut19:
We might want to follow [this](https://devblogs.microsoft.com/oldnewthing/20190916-00/?p=102892) blog?

ramirezdiego:
Are we rescuing this?

zhut19:
We should test this on some other repo first. And since it's not urgent right now, I will put it on hold unless someone else would like to do it.

terliuk:
I would say that advantages of split core is actually more important than having git blame. In the worst case, people can trivially "go" back in time to commits/previous versions and use blame there.

PR #181, opened 2021-07-02
Core splitting

zhut19:
Addressing #170


terliuk:
Great thanks! Just small question, will git blame work in this case? I remember that you posted a solution how to make it, but I am not sure how applicable it is.

zhut19:
> Great thanks! Just small question, will git blame work in this case? I remember that you posted a solution how to make it, but I am not sure how applicable it is.
> 
> Also, it's pretty much copy-paste of old core.py to new files, right?

The blame is working, it's using git mv to rename core.py, then checkout HEAD~ core.py to restore core.py, so now there're two of them.

PR #185, opened 2021-07-08
Fix removed nveto

mzks:
Reopen for #171 rest discussions.


mzks:
I discussed with Diego in private chat, I understand that his request is to add additional cut based on `entry_start` and `entry_stop` for also nVeto.
It has been already here : https://github.com/XENONnT/WFSim/blob/83f0df03dcc2c7e2841134e959be06f4a1a1fe64/wfsim/strax_interface.py#L154

Thus, I think that we can merge without additional lines.
I checked that the `entry_stop` works for also nVeto in my local environment. @ramirezdiego

mzks:
https://github.com/XENONnT/WFSim/pull/171#discussion_r665125310
Original discussions

wfsim/strax_interface.py
```diff
@@ -592,10 +592,9 @@ def get_instructions(self):
...
             self.instructions_nveto, self.nveto_channels, self.nveto_timings =\
                 read_optical(self.config_nveto)
-            # Why epix removes many of the g4ids?
-            # Remove nveto event if no tpc event of the same g4id is found
             if len(self.g4id) > 0:
                 nv_inst_to_keep = (self.instructions_nveto['_last'] - self.instructions_nveto['_first']) >= 0
+
             self.instructions_nveto = self.instructions_nveto[nv_inst_to_keep]
             self.g4id.append(self.instructions_nveto['g4id'])
             log.debug("Epix produced %d instructions in nv" % (len(self.instructions_nveto)))
```
ramirezdiego, line 600: "Epix" is still not true here :)

ramirezdiego (changes requested):
I guess that you do not want to upload the Kr83m notebook, right?

wfsim/strax_interface.py
mzks: Sorry, I cannot understand your comment. Which epix in the code?
Do you mean we need to  keep `# Why epix removes many of the g4ids?` this comment?

mzks:
> I guess that you do not want to upload the Kr83m notebook, right?

Thanks. It may be my mistake of git handling.
Anyway, I remove manually.

wfsim/strax_interface.py
ramirezdiego: It was a minor comment. "Epix produced %d instructions in nv" is not correct, because nVeto entries and epix don't talk to each other.

mzks: OK, I see.

ramirezdiego (approved):
I guess this is ready to merge. The changes are just cosmetic, since the main point was to perform the check you describe above.

PR #188, opened 2021-07-14
Core splitting 2

### LINKED ISSUE #170: Split and refactor core.py

zhut19:
Another attempt of #181 to fix #170


ramirezdiego:
Nice!
image

ramirezdiego (approved):
🤞

PR #190, opened 2021-07-20
Only nveto mc chain

mzks:
**What is the problem / what does the code in this PR do**
When there is only nVeto event in Geant4 file, for example, optical calibration, wfsim took errors related to epix.
This PR add treatments for that.
I added also treatments for epix https://github.com/XENONnT/epix/pull/36

PR #191, opened 2021-07-20
Only nveto fix

mzks:
#190 PR will solve using `targets` option for example,
```
st.set_config(dict(
    neutron_veto=True,
    fax_config_nveto='fax_config_nt_nveto.json',
    ....
    targets=('nveto',),
))
```
We don't need to change epix for that.
It is basically simple solution, but we need to change WFSim code a bit.
This PR is the treatment.
The local test like #190 passed.


petergaemers:
Hi Keita, can you put back the notebook...?

mzks:
#190

PR #194, opened 2021-07-30
Fix optical readout

petergaemers:
In optical_adjustment we split peaks if they are too long, adding a new instruction for the remainder. However there is a mistake causing the split peak to not be split and the remainer to be still added afterwards

This fixes the splitting function

### DIFF
--- a/tests/test_contexts.py
+++ b/tests/test_contexts.py
@@ -17,7 +17,7 @@ def test_nt_context(register=None, context=None):
         return
 
     if context is None:
-        context = straxen.contexts.xenonnt_simulation()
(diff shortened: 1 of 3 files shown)


terliuk:
Thanks Peter! It seems to fix that issue that I observed with range overlaps/duplication after splitting. I am wondering whether it can also explain issues seen by Keita, as it looks more like scrambling rather than duplication in that case, right?

petergaemers:
The bug gave a partial duplication, so this might be what Keita saw before

zhut19 (approved):

PR #229, opened 2021-10-05
Introduce additional parameter to spread SPE distribution

mzks:

For neutron Veto, template pulse and SPE distributions are implemented. However, the generated SPE distributions and template pulses reproduce  `area` of `hitlets_nv` but not perfectly `amplitude` and others.
It might come from WFSim assumption: all pulse can be reproduced sum of SPE template pulse * gain with spread. In fact some pulses spread with vertical axis (time) because of the time spread of charge reaching timings.

I tried to fix the two parameters, `area` and `amplitude` with one additional spread parameter (`additional_spe_spread_charge`), which is sigma of gaussian distribution with unit of charge.

image


mzks:
Related note : https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:nveto:pulse_check1_v2

mzks:
Wiki page : https://xe1t-wiki.lngs.infn.it/doku.php?id=mzks:analysis:wfsim:nveto_photon_area_collection

PR #246, opened 2021-10-30
Add n_pe to truth

petergaemers:
Right now the n_photons field in the truth is the number of detected photons + double pe emission.
This PR changes that to have n_photons be the number of detected photons and n_pe for the number of detected photo electrons (so photons+dpe)


JoranAngevaare:
@petergaemers , this is super useful! Would you think we could also add a small field for the number of photons actually passing the DAQ-thresholds, as in the number of "truely" registered hits?

ershockley:
@petergaemers thanks for this! Can we also save the true sampled gain values?

petergaemers:
Hi guys, I'm sorry I completly missed these comments.

First for Jorans we don't have a number of photons passing the threshold, we have a current which is a bunch of them.

And for Evans point: maybe but it will not be straightforward at all. The main point is straxs fixed length arrays you need to set in advance

JoranAngevaare:
@petergaemers do you think it would make sense to remove the `exclude_dpe_in_truth` option all-together from WFSim after this change?

petergaemers:
Yes I do

PR #260, opened 2021-11-03
S2aft

petergaemers:
Use a skewed gaussian for the s2 area fraction top distribution


ramirezdiego:
@petergaemers @zhut19 This was merged but the corresponding config file PR is pending: https://github.com/XENONnT/private_nt_aux_files/pull/112/files

It would also be nice to document the reason for this change somewhere in a note or in the wFSim summary page, for reference.

PR #262, opened 2021-11-10
Fix nVeto dtypes handling in RawRecordsFromMcChain

ramirezdiego:

Loop over nVeto data types only when `nveto` is one of the targets to simulate, otherwise we run into errors when using the mc_chain context with nVeto set to false: we're requesting `result_nv` without having defined it earlier (because `nveto` it's not in targets), while the `RawRecordsFromMcChain` plugin provides with all data types for nVeto and TPC.


HenningSE (approved):
I could deactivate the nv simulation using this branch and no errors popped up. So I think your fix works fine @ramirezdiego.

PR #264, opened 2021-11-15
Exclude DPE from n_photon in truth

JosephJHowlett:
Currently `n_photon` in the truth output of WFSim is not actually a number of photons, but a number of photoelectrons emitted from the PMT photocathodes. This is fine as an option, but I think it should not be the default since it isn't quite accurate. After this PR, `n_photon` would refer to the number of photons detected by PMTs (making at least one PE at the photocathode).

### DIFF
--- a/wfsim/core/rawdata.py
+++ b/wfsim/core/rawdata.py
@@ -343,7 +343,7 @@ class RawData(object):
                 * self.config['sample_duration']
 
         channels = getattr(pulse, '_photon_channels', [])
-        if self.config.get('exclude_dpe_in_truth', False):
+        if self.config.get('exclude_dpe_in_truth', True):
             n_dpe = n_dpe_bot = 0
         else:
             n_dpe = getattr(pulse, '_n_double_pe', 0)


ershockley (approved):
Thanks @JosephJHowlett! I think this works. Maybe just one comment: is there ever a case when we would want to include DPE in truth? I would prefer to remove this option entirely and save in the truth the photons_detected _and_ the number of 'DPE' events. 

But this is easier solution for now for sure.

JoranAngevaare (approved):
Thanks Joey, in addition to Evans comment which I fully support I would argue to also add such a hidden configuration in the fax config file

ramirezdiego:
> 
> 
> Thanks Joey, in addition to Evans comment which I fully support I would argue to also add such a hidden configuration in the fax config file

Agreed!

petergaemers:
Hi Joey,
Thanks for taking initiative. I think just storing both is better like here:
https://github.com/XENONnT/WFSim/pull/246

JosephJHowlett:
Great! I missed it, sorry @petergaemers ... Closing now and lets merge the other then

PR #265, opened 2021-11-16
add infer_type=False to all strax.Options

jmosbacher:
Due to a bug fix in strax PR #569 many config options in WFSim will now produce warnings. This PR suppresses these warning by explicitly setting the infer_dtype to False wherever the type of the config Option would have been inferred from the default value in the future.
This breaks backward compatibility with older strax versions since the infer_dtype keyword did not exist before, so I have pinned the strax version in the requirements to be above the next version.

A better alternative to this PR would be to set the type explicitly for all options, but in case you dont want to commit to a specific type for all of your plugin options this PR will just disable the auto inference of types from the default values.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-strax
+strax>=1.1.2
 utilix>=0.5.3
 straxen>=0.19.1
 uproot>=4.0.0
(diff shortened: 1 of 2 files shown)


JoranAngevaare (approved):
Thanks Yossi!

PR #269, opened 2021-11-22
Recompute lce on the run

zhut19:
As @terliuk found out, the turned-off PMT in LCE must match the ones in the `to_pe` model. This is especially true for S2s, LCE under 'OFF' PMTs is smaller, while the relative probabilities in the 'ON' PMTs are higher.


terliuk (approved):
Seems to work as expected, at least I do not see any features in the patterns that I used to have before.

PR #283, opened 2021-12-06
SPE spec bug fix, adding trigger truth

zhut19:
  - Fixing a bug in the SPE spectrum, we were only using channel 0 for nT.
  - Also adding `trigger` truth to the truth values, including 
    - `n_photon_trigger`: number of photons that should pass the ADC trigger assuming no noise fluctuation; this also considers the possibility of one photon generating 2 PE and thus having a higher chance of passing ADC threshold. 
    - `n_pe_trigger`: number of photons + dpe that should pass the ADC trigger assuming no noise fluctuation
  - Adding a new field `raw_area`: Sum of the gain per PE triggered
  - Those three fields has a copy of bottom only values


zhut19:
Hi @JoranAngevaare, I am adding the required fields including `n_photon_trigger` to the truth output. But also realized there's a pretty obvious bug that for nT, we were using only the channel[0] single photon electron spectrum for all channels.

wfsim/core/pulse.py
```diff
@@ -196,6 +210,41 @@ def uniform_to_pe_arr(self, p, channel=0):
...
+                  photon_gains,
+                  dt,
+                  channel,
+                  n_double_pe,
+                 ):
+
+        if channel in self.config['turned_off_pmts']:
+            return
+
+        if str(channel) in self.config.get('special_thresholds', {}):
```
FaroutYLq, line 224: Sorry can you remind me what is `special_thresholds`? ~~And where is the rounding that makes us want to do the `-0.5`?~~ No worries I figured out. The rounding might be from digitization

wfsim/core/pulse.py
```diff
@@ -196,6 +210,41 @@ def uniform_to_pe_arr(self, p, channel=0):
...
+            threshold = self.config['special_thresholds'][str(channel)] - 0.5
+        else:
+            threshold = self.config['zle_threshold'] - 0.5
+
+        reminder = (photon_timings % dt).astype(int)
+        above_threshold = photon_gains * self.current_max[reminder] * self.current_2_adc > threshold
+        trigger_photon = np.sum(above_threshold)
+        trigger_dpe = np.sum(above_threshold[:n_double_pe])
+        raw_area = np.sum(photon_gains[above_threshold]) / self.config['gains'][channel]
```
FaroutYLq, line 233: Just to clarify. Here `raw_area` means the area of photon signal which is high enough to pass DAQ threshold. Is there a something like `true_area = np.sum(photon_gains) / self.config['gains'][channel]` already? I believe it would also be useful for some bias study in `wfsim`, though I don't know how at this moment.

wfsim/core/pulse.py
```diff
@@ -196,6 +210,41 @@ def uniform_to_pe_arr(self, p, channel=0):
...
+
+        self._n_photon += len(photon_timings)
+        self._n_pe += len(photon_timings) + n_double_pe
+        self._n_photon_trigger += trigger_photon
+        self._n_pe_trigger += trigger_photon + trigger_dpe
+        self._raw_area += raw_area
+
+        if channel in self.config['channels_bottom']:
```
FaroutYLq, line 241: Very nice to see we have separated array information. Andrii might be very happy: )

FaroutYLq (approved):
Thanks a lot for this PR! This is really helpful.

wfsim/core/pulse.py
zhut19: Special channels are those not set to 15 ADC, usually more to reduce noise triggers.

wfsim/core/pulse.py
zhut19: The true gains are set to 1 PE

wfsim/core/pulse.py
```diff
@@ -69,24 +77,27 @@ def __call__(self, *args):
             if '_photon_gains' not in self.__dict__:
 
                 _channel_photon_gains = self.config['gains'][channel] \
-                    * self.uniform_to_pe_arr(np.random.random(len(_channel_photon_timings)))
+                    * self.uniform_to_pe_arr(np.random.random(len(_channel_photon_timings)), channel)
```
JoranAngevaare, line 80: good catch!

wfsim/core/pulse.py
```diff
@@ -246,6 +292,17 @@ def add_current(photon_timings,
...
             pmt_current_templates[reminder] * gain_total
 
+    @staticmethod
+    def hit_above_threshold(photon_timings,
+                            photon_gains,
+                            dt,
+                            pmt_current_templates,
+                            config,):
+
+        reminder = int(photon_timings % dt)
```
JoranAngevaare, line 302: This is not used, right?

wfsim/core/pulse.py
```diff
@@ -199,6 +210,41 @@ def uniform_to_pe_arr(self, p, channel=0):
...
+        if channel in self.config['turned_off_pmts']:
+            return
+
+        if str(channel) in self.config.get('special_thresholds', {}):
+            threshold = self.config['special_thresholds'][str(channel)] - 0.5
+        else:
+            threshold = self.config['zle_threshold'] - 0.5
+
+        reminder = (photon_timings % dt).astype(int)
+        above_threshold = photon_gains * self.current_max[reminder] * self.current_2_adc > threshold
```
JoranAngevaare, line 230: Sorry I might be a bit slow, but I must admit that I don't fully grasp what you do here, could you help me a bit?

wfsim/core/pulse.py
```diff
@@ -199,6 +210,41 @@ def uniform_to_pe_arr(self, p, channel=0):
...
         return self.__uniform_to_pe_arr[channel, indices]
 
+    def add_truth(self,
+                  photon_timings,
+                  photon_gains,
+                  dt,
+                  channel,
+                  n_double_pe,
+                 ):
+
```
JoranAngevaare, line 220:
```suggestion
                  ):
        """Add required information to the fields used for the truth information"""
```

wfsim/core/rawdata.py
```diff
@@ -342,19 +342,10 @@ def get_truth(self, instruction, truth_buffer):
...
-        # this turned_off guy, check how this works with a config['turned_off_guys']
-        channels_bottom = list(
-            set(self.config['channels_bottom']).difference(getattr(pulse, 'turned_off_pmts', [])))
-        tb['n_photon_bottom'] = (
-            np.sum(np.isin(channels, channels_bottom)))
-        tb['n_pe_bottom'] = tb['n_photon_bottom'] + n_dpe_bot
+        # Copy single valued fields directly from pulse class
+        for field in ['n_pe', 'n_pe_trigger', 'n_photon', 'n_photon_trigger', 'raw_area']:
+            for suffix in ['', '_bottom']:
+                tb[field+suffix] = getattr(pulse, '_' + field + suffix, 0)
```
JoranAngevaare, line 348: Nice

JoranAngevaare (commented):
Thanks Tianyu for this PR! Much appreciated. One piece of the code I don't fully follow, can you help me a bit?

wfsim/core/pulse.py
zhut19: Ah, good catch, forgot to delete this

wfsim/core/pulse.py
zhut19: Yeah, it's quite confusing.
  - Current max is the highest value in the SPE pulse model given a reminder [0-9] ns (offset from 10 ns sample time). The SPE pulse model sums up to 0.1 (pe/ns), and the peak is ~0.03 (pe/ns) 
  - Multiply this by the sampled gain of each photon, we get (electron/ns)
  - The current_2_adc take into account n_electron -> current -> voltage -> amplification -> digitization, converting the electron/ns to adc value.

wfsim/core/pulse.py
FaroutYLq: Ohoh so it's already in wfsim? Thanks a lot for clarification.

wfsim/core/pulse.py
JoranAngevaare: Thanks a lot, this is instructive indeed 👍

JoranAngevaare: I'll add it to the code as a comment if you don't mind? 

I think there is also a small typo, it should be `remainder` right?

JoranAngevaare:
I tested it and it really works great, thanks again Tianyu!

PR #285, opened 2021-12-14
Fdc quick fix

zhut19:
FDC map coordinate system is in drift time and should be rescaled for interpolating x-y-z input.


ftoschi:
@zhut19 A question for this one: if we do this, don't we lose compatibility with (x,y,z) maps from 1T? Do you think it would make sense to have something deciding based on the average of the 3rd component? If positive it must be drift time, so scale, otherwise it is a z coordinate?

zhut19:
@ftoschi This part is added under the 'xenonnt' detector condition, so I would say 1T is alright.

ftoschi:
Oh yes, thanks, I did not check carefully enough!

ftoschi (approved):

ftoschi:
I do not really have time now to test it, but @zhut19 is right as the FDC map should be scaled.

PR #286, opened 2021-12-21
Use straxen#833

JoranAngevaare:
Use https://github.com/XENONnT/straxen/issues/833 to fix line endings from windows


.gitattributes
```diff
@@ -0,0 +1,9 @@
+# https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/#a-simple-gitattributes-config
+# We'll let Git's auto-detection algorithm infer if a file is text. If it is,
+# enforce LF line endings regardless of OS or git configurations.
+* text=auto eol=lf
+
+# Isolate binary files in case the auto-detection algorithm fails and
+# marks them as text files (which could brick them).
+*.{png,jpg,jpeg,gif,webp,woff,woff2} binary
```
kdund, line 8: I'm not aware of any, but do we have any other binary formats in straxen? (npz etc?) I think no.

JoranAngevaare: Hi Knut, In principle we shouldn't even have binary files in the code itself so I also don't think it's an issue. I just copy pasted these few from the link.

Thanks!

zhut19 (approved):

ramirezdiego (approved):

Issue #287, opened 2021-12-24
S1/S2 grouping in truth


PavelKavrigin:
Currently the information written to the truth output of WFSim is reduced: 

All S1s within 100 ns separation are merged:
https://github.com/XENONnT/WFSim/blob/0d74d97f1e534f0383fdd2791eca5be8e26831b0/wfsim/core/rawdata.py#L109

All S2s within 2 mm separation are merged:
https://github.com/XENONnT/WFSim/blob/0d74d97f1e534f0383fdd2791eca5be8e26831b0/wfsim/core/rawdata.py#L114

As a result, the truth cannot be used for any analysis which requires a knowledge of true multi-scatters in the instructions since this information is lost if the interactions happen close to each other. The most simple fix would be just to change these thresholds, e.g. 100 ns to 0.1 ns, 2 mm to 1 um. In case if we want to keep a possibility to reduce the size of the truth output, we will need to add some flag to the configuration to specify whether we want to merge instructions in the truth or not.

JoranAngevaare:
Seems to be part of https://github.com/XENONnT/WFSim/pull/125, specifically this commit: https://github.com/XENONnT/WFSim/pull/125/commits/8de52e04b56ad204f4e92d37432d53b4031c771e

In my view, your proposal makes a lot of sense

zhut19:
Fixed in #288

PR #288, opened 2021-12-29
Save full truth output

PavelKavrigin:
This PR addresses the issue [https://github.com/XENONnT/WFSim/issues/287](https://github.com/XENONnT/WFSim/issues/287). User can set the flag **save_full_truth** in **fax_config_override** to True (record the whole truth output) or False (merge S1/S2 in the truth as it is currently done in the master branch). The default value is True.


zhut19 (approved):

ramirezdiego:
@zhut19 Was this "clustering" set to save space? I remember we discussed it but forgot the reason to achieve this compromise. Just asking in case we want to set the default to false.

zhut19:
> Was this "clustering" set to save space?

I am not sure I was in this discussion, I just assumed the point of this pr @PavelKavrigin is that for the full chain one might want the full truth output. Do you think you would like 'false' by default? While I agree the full chain file might contain too many primary instructions, the truth buffer size now should be able to handle it.

ramirezdiego:
I meant the discussion right before this 

> > Was this "clustering" set to save space?
> 
> I am not sure I was in this discussion

I meant the discussion that lead to the current behavior (i.e., before Pavel's PR). I remember we agreed on setting this 2 mm grouping for S2s and I forgot the reason behind. But if we agree on leaving `true` as default we can just move on.

PR #305, opened 2022-01-31
S2 optical prop

KaraMelih:
Samples optical propagation delays in the gas phase from a G4 simulated data.
The S2 pulse workflow is a bit changed; photon channels sampling is now independent from photon times. This was needed so that photon time delays can be added based on whether they are detected at top/bottom arrays.
!To add this effect it uses a local aux file for now, the file will be added to private_nt_aux repo!

**WFSim needs you:**

*Pay attention:*

### DIFF
--- a/wfsim/core/s2.py
+++ b/wfsim/core/s2.py
@@ -270,8 +270,8 @@ class S2(Pulse):
         assert len(n_photons) == len(xy), 'Input number of n_electron should have same length as positions'
         assert len(resource.s2_luminescence['t'].shape) == 2, 'Timing data is expected to have D2'
 
-        if confxy:
-            distance = np.random.uniform(-0.01, 0.01, len(xy))
+        if type(confxy)==float:
+            distance = np.random.uniform(-confxy, confxy, len(xy))
         else:
             tilt = config.get('anode_xaxis_angle', np.pi / 4)
             pitch = config.get('anode_pitch', 0.5)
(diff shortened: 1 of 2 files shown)


JYangQi00:
Hi Melih, this looks good, but how do you propose we change the config to include 'confxy'?

wfsim/core/s2.py
```diff
@@ -332,42 +330,96 @@ def photon_timings(t, n_electron, z, xy, sc_gain, config, resource, phase):
...
+        :param config: dict of the wfsim config
+        :param resource: instance of the resource class
+        """
+        # Luminescence Timings
+        if 'simple' in config['s2_luminescence_model']:
+            photon_timings = S2.luminescence_timings_simple(xy, n_photons_per_xy,
+                                                             config=config,
+                                                             resource=resource)
+        elif 'garfield' in config['s2_luminescence_model']:
+            confxy = True if 'confine distance' in config['s2_luminescence_model'] else False
```
KaraMelih, line 390: @JYangQi00 Hi Jianyang currently, I added it here. So while comparing the two, I simulate once setting `{'s2_luminescence_model':'garfield'}` and once with `{'s2_luminescence_model':'garfield confine distance'}` so both `confxy=True` and `confxy=False` are callable

KaraMelih: and [here](https://github.com/XENONnT/analysiscode/blob/master/SE_WFSim_shape/data-WFSim_comparison.ipynb) I uploaded a notebook, using this updates. I'm trying to compare the data vs WFSim with several configs

JYangQi00 (approved):
Okay, I see, looks good!

PR #316, opened 2022-02-01
up py version

JoranAngevaare:
Follow AxFoundation/strax#636, XENONnT/straxen#906, let's drop support for older python versions.

For nestpy, see the discussion in https://github.com/NESTCollaboration/nestpy/pull/80
Also see https://github.com/NESTCollaboration/nestpy/pull/80

PR #317, opened 2022-02-04
reduce warnings

JoranAngevaare:
Currently have about 100 warnings, let's reduce that

afbeelding

similar to https://github.com/XENONnT/straxen/pull/915

### After
afbeelding


JoranAngevaare:
requires https://github.com/XENONnT/WFSim/pull/316

PR #320, opened 2022-02-11
Jianyang wfsim tests

KaraMelih:
Single Electron shape data-simulation matching improvements
[Jianyang's note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:jianyang:wfsim_se_comparison)
[Melih's note](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:kara_se_shape_wfsim_matching)
**WFSim needs you:**

*Pay attention:*

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 0.5.12
+current_version = 0.5.13
 files = setup.py wfsim/__init__.py
 commit = True
 tag = True
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,10 @@
+v0.5.13 / 2022-01-17
+====================
+* Save full truth output (#288)
+* Bump strax(en) (#303)
+* Use straxen#833 (#286)
+* Integrated testing and requirements (#304, #289, #290, #291, #293, #294 #295, #296, #297)
+
 v0.5.12 / 2021-12-10
 ====================
 * Rename s2_aft_sigma config (#284)
--- a/extra_requirements/requirements-tests.txt
+++ b/extra_requirements/requirements-tests.txt
@@ -1,6 +1,6 @@
 # Test compatibility with these most often:
 strax==1.1.5
-straxen==1.2.4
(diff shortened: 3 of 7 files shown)


JYangQi00 (approved):

JYangQi00:
Hi Tianyu, the note here:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:jianyang:wfsim_se_comparison
already assumes a liquid level of 5.1mm

files/XENONnT_wfsim_config.json
```diff
@@ -8,6 +8,7 @@
 {
     // Configuration flags
     "s1_model_type": "simple",
+    "s2_time_model": "s2_time_spread around zero",
```
KaraMelih, line 11: Isn't it better if the default is `'optical_propagation'` ?
We also set that for the sr0 config [here](https://github.com/XENONnT/private_nt_aux_files/blob/02edbf6ba2e977aabc5e412d8558864ddd9497d4/sim_files/fax_config_nt_sr0_v0.json#L25)

zhut19: Hi, This part of [the test](https://github.com/XENONnT/WFSim/blob/19dde4b1a7b833cf41c823c0a4fecfad5f386b14/tests/test_wfsim.py#L62) won't be loading any data from our database, so we can't use optical propagation. While the [test_sim_nT_advanced](https://github.com/XENONnT/WFSim/blob/19dde4b1a7b833cf41c823c0a4fecfad5f386b14/tests/test_wfsim.py#L104) will use sr0 config and thus also test the optical map. And this way we can also maximize test coverage.

zhut19:
@KaraMelih I also uploaded the optical map in a format that can be directly used by straxen in https://github.com/XENONnT/private_nt_aux_files/pull/167. Can you help double-check it's consistent with your result? Thanks! Also, that pr has to be merged before the tests here pass since the test on that part fetches the file through our MongoDB, which only updates with the master branch.

PR #321, opened 2022-02-16
PMT afterpulse not sorted by channel

zhut19:
One of the optimizations in #15 on the `pulse` class created this bug, that PMT afterpulse are ordered in channel number instead of from the channel created them.

https://github.com/XENONnT/WFSim/blob/c50a7a2b7c67b37755e1d073683f902c9424f111/wfsim/core/pulse.py#L67

To resolve this, the returned values need to be sorted by the channel numbers.

### DIFF
--- a/wfsim/core/afterpulse.py
+++ b/wfsim/core/afterpulse.py
@@ -230,7 +230,8 @@ class PMT_Afterpulse(Pulse):
             _photon_amplitude = np.hstack(_photon_amplitude)
             _photon_gains = np.array(config['gains'])[_photon_channels] * _photon_amplitude
 
-            return _photon_timings, _photon_channels, _photon_gains
+            sorted_index = np.argsort(_photon_channels)
+            return _photon_timings[sorted_index], _photon_channels[sorted_index], _photon_gains[sorted_index]
 
         else:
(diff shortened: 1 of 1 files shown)


terliuk:
Thanks! Can we test it using the same notebook as @hoetzsch ? It would be nice to verify if AP problem is resolved by it.

hoetzsch:
Hey @skazama, I'm working on it right now. There were some problems with getting enough statistics, hence the delay. The notebook is here: https://github.com/XENONnT/analysiscode/blob/master/wfsim/test_pmtafterpulses_wfsim.ipynb

hoetzsch (approved):
Hey,
I just updated the [notebook](https://github.com/XENONnT/analysiscode/blob/master/wfsim/test_pmtafterpulses_wfsim.ipynb) for the afterpulse tests, and below is the new afterpulse delay time vs. channel plot, which now shows correct ordering.
However I can't reproduce my afterpulse rates from data with WFsim, I get rates from WFsim that are systematically lower by about 25%. Not sure yet why (could be the input or pile-up effects), but it's likely not related to the channel mismatch, so I think this can be merged.

aps_wfsim_channelorderfix

PR #338, opened 2022-03-24
Fix CIV and average drift velocity implementation

terliuk:

 - fixed CIV implementation
 - fixed average drift velocity map implementation with normalization for cathode-gate drift time at R=0
 - removed potentially dangerous code duplication

### DIFF
--- a/wfsim/core/s2.py
+++ b/wfsim/core/s2.py
@@ -71,6 +71,27 @@ class S2(Pulse):
         super().__call__()
 
     @staticmethod
+    def get_avg_drift_velocity(z, xy, config, resource):
+        """Calculate s2 drift time mean and spread
+
+        :param positions: 1d array of z (floats)
+        :param xy: 2d array of xy positions (floats)
+        :param config: dict with wfsim config
+        :param resource: instance of the resource class
+
+        returns array of floats corresponding to average drift velocities from given point to the gate
+        """
+        drift_v_LXe=None
+        if config['enable_field_dependencies']['drift_speed_map']:
+            drift_v_LXe = resource.field_dependencies_map(z, xy, map_name='drift_speed_map')  # mm/µs
+            drift_v_LXe *= 1e-4  # cm/ns
+            drift_v_LXe *= resource.drift_velocity_scaling
+            drift_v_LXe
+        else:
+            drift_v_LXe=config['drift_velocity_liquid']
+        return(drift_v_LXe)
+
+    @staticmethod
     def get_s2_drift_time_params(z_obs, positions, config, resource):
         """Calculate s2 drift time mean and spread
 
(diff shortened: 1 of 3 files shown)


wfsim/core/s2.py
```diff
@@ -43,15 +43,16 @@ def __call__(self, instruction):
         else:
             z_obs, positions = z, np.array([x, y]).T
 
+        n_electron = self.get_electron_yield(n_electron=n_electron,
```
ftoschi, line 46: Could you change also the docstring of this function where it says "observed"? Also, for readability it would be better to change from z_obs to z

wfsim/core/s2.py
```diff
@@ -139,18 +155,19 @@ def get_electron_yield(n_electron, positions, z_obs, config, resource):
         # Average drift time of the electrons
         drift_time_mean, drift_time_spread = S2.get_s2_drift_time_params(z_obs, positions, config, resource)
 
-        # Absorb electrons during the drift
+        # Absorb electrons during the drift, clipping to avoid artefacts of maps
+        drift_time_mean=np.clip(drift_time_mean, a_min=0, a_max=np.inf)
```
ftoschi, line 159: Can you just change in the function from `_drift_time_mean` to `drift_time_mean` and avoid the clipping here? I am not sure now why I included it, I do not think such a situation can show up (or at least I forgot if it did), but in this way we do not have too much stuff out of the function

wfsim/core/s2.py
```diff
@@ -139,18 +155,19 @@ def get_electron_yield(n_electron, positions, z_obs, config, resource):
...
         cy = config['electron_extraction_yield'] * electron_lifetime_correction
 
         # Remove electrons in insensitive volume
         if config['enable_field_dependencies']['survival_probability_map']:
-            survival_probability = resource.field_dependencies_map(z_obs, positions, map_name='survival_probability_map')
-            cy *= survival_probability
-
-        # why are there cy greater than 1? We should check this
-        cy = np.clip(cy, a_min = 0, a_max = 1)
+            p_surv = resource.field_dependencies_map(z_obs, positions, map_name='survival_probability_map')
```
ftoschi, line 166: This is actually quite a mistake, this should really have the real (x,y,z), I am wondering now how it is that we actually saw a decrease in the rate with z

wfsim/core/s2.py
```diff
@@ -139,18 +155,19 @@ def get_electron_yield(n_electron, positions, z_obs, config, resource):
...
+            if np.any(p_surv<0) or np.any(p_surv>1):
```
ftoschi, line 167: Is it necessary this if statement? Clipping simply does nothing if this is false

ftoschi (changes requested):

wfsim/core/s2.py
```diff
@@ -383,13 +400,8 @@ def s2_pattern_map_diffuse(n_electron, z, xy, config, resource):
...
         """
         assert all(z < 0), 'All S2 in liquid should have z < 0'
-        
-        if config['enable_field_dependencies']['drift_speed_map']:
-            drift_velocity_liquid = resource.field_dependencies_map(z, xy, map_name='drift_speed_map')  # mm/µs
-            drift_velocity_liquid *= 1e-4  # cm/ns
-        else:
-            drift_velocity_liquid = config['drift_velocity_liquid']
-            
+        drift_velocity_liquid=S2.get_avg_drift_velocity(z, xy, config, resource)
```
ftoschi, line 403: This should use the real xy, but the rest of timing estimation should use the observed one

ftoschi:
Where is the code duplication?

wfsim/core/s2.py
terliuk: Is it like that? when I plot maps it looks like the intention is actually (x,y,z) true that gives average velocity.

ftoschi: For the drift speed map you need the true (interaction) position, so this function should take it as it already does in the code. Then when you are sampling from maps such as the GARFIELD timing, sagging or S2 pattern map you should be using the field distorted position. Actually, you should be using the position of electrons at the liquid interface which is the field distorted one + the gate focusing effect, but I think this is within the posrec resolution and it is not a major probelm

ftoschi (approved):

PR #340, opened 2022-03-28
Double PMT ap probability for photons emitting double PE.

zhut19:
The double photo-electron in PMTs also contributes to generating PMT afterpulse, in this pr, we store a boolean array noting photons emitting dpe, in the last step of generating S1s and S2s. While we double the PMT AP probability for those photons when generating PMT afterpulses.

### DIFF
--- a/wfsim/core/afterpulse.py
+++ b/wfsim/core/afterpulse.py
@@ -194,7 +194,7 @@ class PMT_Afterpulse(Pulse):
                 log.warning(f'PMT after pulse probability is {prob} larger than 0.5?')
 
             # Double the probability for those photon emitting dpe
-            cdf_max[signal_pulse._photon_is_dpe] *= 2
+            rU0[signal_pulse._photon_is_dpe] /= 2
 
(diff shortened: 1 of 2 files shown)


ramirezdiego (dismissed):

hoetzsch:
Hey @zhut19,

thanks for the fix, now the population with the additional delay is gone!
The first plot below shows the ratio of the He afterpulse rates I get from the wfsim data after and before the DPE fix, and it turns out to be close to the dpe probability, so this seems to work well.

wfsimaps_ratioafterbeforedpe

I still can't quite reproduce the afterpulse rates from the input file though, but the discrepancy has decreased significantly with the DPE addition, from -33% to -18%, shown in the second plot. Still not sure where this comes from, could be pile-up. But I don't think it's related to the DPE thing.

wfsimaps_comptoinput

hoetzsch (approved):

terliuk (approved):
It seems like it does what this PR is supposed to do, so let's merge it and figure out why there is still remaining mismatch separately.

ramirezdiego (approved):
Agreed

PR #361, opened 2022-04-08
Change default compressor

JoranAngevaare:
I just noticed - to my surprise - that blosc is still the default compressor, whereas  we can be about ~2x squeezer than that at almost no performance loss:
https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:ivyli:lossless_compression

### DIFF
--- a/wfsim/strax_interface.py
+++ b/wfsim/strax_interface.py
@@ -474,6 +474,7 @@ class ChunkRawRecords(object):
                       "generation of the instructions"),
 )
 class SimulatorPlugin(strax.Plugin):
+    compressor = 'zstd'
     depends_on = tuple()
 
     # Cannot arbitrarily rechunk records inside events


JoranAngevaare:
@terliuk @ramirezdiego @ershockley 
I think we should go ahead and merge this

It's tested here:
https://github.com/XENONnT/WFSim/blob/77868635d39371cdec7d5abc94735b5cccbbb99a/tests/test_wfsim.py

Additionally I've tested re-compressing the data using zstd, it lead to a compression of `3.6` whereas `blosc` does `1.9`.

Downside is that loading would be ~twice as slow, but you rarely really need to go down to raw-records.

ramirezdiego (approved):
Thanks!

ershockley:
Sorry missed this one somehow!

PR #363, opened 2022-04-28
XY-dependent SE gain and extraction efficiency

JYangQi00:

This PR adds: Sampling from the data-driven S2-xy correction and single electron gain maps in order to draw the proper xy dependent extraction efficiency and single electron gain, a gas-gap map from garfield so that the single electron waveform shape can be matched at different positions in the detector, and a new garfield electroluminescence map which samples only from below the anode but at different gas gaps.

**WFSim needs you:**

*Pay attention:*


zhut19:
@JYangQi00 this looks good, however, can you make it work with the older version configs? Now it seems that one can not run the simulation without having a XENONnT SE gain map.

JYangQi00:
> @JYangQi00 this looks good, however, can you make it work with the older version configs? Now it seems that one can not run the simulation without having a XENONnT SE gain map.

Hi Tianyu, it seems that I might need to make a pull request to the private nt aux files first, since the structure of this plugin may be quite different. But I nonetheless attempted to allow for compatibility with the previous version (though it's a bit clunky I must admit)

KaraMelih:
could you please link your wiki note and the private not aux PR/branch here?

wfsim/core/s2.py
```diff
@@ -189,15 +189,27 @@ def get_s2_light_yield(positions, config, resource):
...
         returns array of floats (mean expectation) 
         """
-        sc_gain = resource.s2_correction_map(positions)
+        
+        if ('se_gain_from_map' not in config):
+            sc_gain = resource.se_gain_map(positions)
+        else:
+            if config['se_gain_from_map'] == False:
+                sc_gain = resource.se_gain_map(positions)
+            sc_gain = resource.se_gain_map(positions)
```
KaraMelih, line 198: in this conditional all conditions return the same things, don't they?
also, what is `sc` stand for? this is more for me to understand

KaraMelih: I guess it should be;
```python
       # calculate it from MC pattern map directly if no "se_gain_map" is given 
        if ('se_gain_from_map' not in config):
            sc_gain = resource.s2_correction_map(positions)
        else:
            if config['se_gain_from_map'] == False:
                sc_gain = resource.s2_correction_map(positions)
            else:
                sc_gain = resource.se_gain_map(positions)
```

wfsim/core/s2.py
```diff
@@ -221,7 +233,7 @@ def get_electron_yield(n_electron, xy_int, z_int, config, resource):
         # Absorb electrons during the drift
         electron_lifetime_correction = np.exp(- 1 * drift_time_mean /
                                               config['electron_lifetime_liquid'])
-        cy = config['electron_extraction_yield'] * electron_lifetime_correction
+        cy = config['g2_mean']*resource.s2_correction_map(xy_int)*electron_lifetime_correction/resource.se_gain_map(xy_int)
```
KaraMelih, line 236: I could not trace back to this `'g2_mean'` is it supposed to be in fax config on your branch?

wfsim/load_resource.py
```diff
@@ -259,12 +268,13 @@ def __init__(self, config=None):
...
-                s2_luminescence_map = straxen.get_resource(files['s2_luminescence'], fmt='npy_pickle')['arr_0']
-                # Get directly the map for the simulated level
-                liquid_level_available = np.unique(s2_luminescence_map['ll'])  # available levels (cm)
-                liquid_level = config['gate_to_anode_distance'] - config['elr_gas_gap_length']  # cm
-                liquid_level = min(liquid_level_available, key=lambda x: abs(x - liquid_level))
-                self.s2_luminescence = s2_luminescence_map[s2_luminescence_map['ll'] == liquid_level]
+                s2_luminescence_map = straxen.get_resource(files['s2_luminescence'], fmt='npy')
+                self.s2_luminescence = s2_luminescence_map
+            
+            if 'garfield_gas_gap' in config.get('s2_luminescence_model', ''):
```
KaraMelih, line 274: I would put this if condition first, and go with `elif 'garfield' in  config...` because when you have `'garfield_gas_gap'` both conditions are correct

wfsim/core/s2.py
JYangQi00: Indeed it is. I just changed my fax config branch to match this. Thanks for pointing it out, I was simply using the fax override to do this before

wfsim/load_resource.py
```diff
@@ -230,11 +233,17 @@ def __init__(self, config=None):
                 self.uniform_to_ele_ap = straxen.get_resource(files['ele_ap_pdfs'], fmt='pkl.gz')
 
         elif config.get('detector', 'XENONnT') == 'XENONnT':
+            
+            print(files)
```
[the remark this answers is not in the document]
JYangQi00: Yes I do, just removed. Thanks!

wfsim/load_resource.py
JYangQi00: Sure, I agree

KaraMelih:
Thanks!
I guess these tests;

> FAILED tests/test_wfsim.py::test_sim_1T - KeyError: 'g2_mean'
> FAILED tests/test_wfsim.py::test_sim_nT_basics - KeyError: 'g2_mean'
> FAILED tests/test_wfsim.py::test_sim_nT_advanced - KeyError: 'g2_mean'
> FAILED tests/test_wfsim.py::test_sim_mc_chain - KeyError: 'g2_mean'`

fail because that key does not exist in the master branch's fax config. 
I also just removed an unused parameter

wfsim/load_resource.py
```diff
@@ -89,10 +89,13 @@ def config_to_file(config):
                 's1_pattern_map': 'XENONnT_s1_xyz_patterns_LCE_corrected_qes_MCva43fa9b_wires.pkl',
                 's1_lce_correction_map': 'XENONnT_s1_xyz_LCE_corrected_qes_MCva43fa9b_wires.json.gz',
                 's2_pattern_map': 'XENONnT_s2_xy_patterns_LCE_corrected_qes_MCva43fa9b_wires.pkl',
-                's2_correction_map': 'XENONnT_s2_xy_correction_corrected_qes_MCva43fa9b_wires.json.gz',
+                's2_correction_map': 'XENONnT_s2_xy_map_v4_210503_mlp_3_in_1_iterated.json',
+                'se_gain_map': 'XENONnT_se_xy_map_v1_mlp.json',
                 'photon_ap_cdfs': 'XENONnT_pmt_afterpulse_config_012605.json.gz',
                 's2_luminescence': 'XENONnT_GARFIELD_B1d5n_C30n_G1n_A6d5p_T1d5n_PMTs1d5n_FSR0d95n.npz',
+                "s2_luminescence_gg": "/dali/lgrandi/jyangqi/electroluminescence_timings/garfield_timing_map_gas_gap.npy",
```
KaraMelih, line 96: I think you can correct the path here and in line 98, since you upload them to private_nt repo right?

JYangQi00: Yeah, I was initially going to keep them there for the time being since this way I could actually use it to run tests. But you're right, I should change it assuming that the private_nt_aux_files PR will go through

wfsim/core/s2.py
```diff
@@ -221,7 +231,9 @@ def get_electron_yield(n_electron, xy_int, z_int, config, resource):
         # Absorb electrons during the drift
         electron_lifetime_correction = np.exp(- 1 * drift_time_mean /
                                               config['electron_lifetime_liquid'])
-        cy = config['electron_extraction_yield'] * electron_lifetime_correction
+        cy = config['g2_mean']*resource.s2_correction_map(xy_int)*\
```
ftoschi, line 234: The S2 correction map and the gain are already included in the S2 light yield, isn't including them here as well equivalent to apply the correction twice?

wfsim/load_resource.py
```diff
@@ -276,13 +281,15 @@ def __init__(self, config=None):
...
-                liquid_level = config['gate_to_anode_distance'] - config['elr_gas_gap_length']  # cm
-                liquid_level = min(liquid_level_available, key=lambda x: abs(x - liquid_level))
-                self.s2_luminescence = s2_luminescence_map[s2_luminescence_map['ll'] == liquid_level]
+            if 'garfield_gas_gap' in config.get('s2_luminescence_model', ''):
+                s2_luminescence_map = straxen.get_resource(files['s2_luminescence_gg'], fmt='npy')
+                self.s2_luminescence = s2_luminescence_map
+                self.garfield_gas_gap_map = make_map(files['garfield_gas_gap_map'], fmt = 'json')
+
+            elif 'garfield' in config.get('s2_luminescence_model', ''):
+                s2_luminescence_map = straxen.get_resource(files['s2_luminescence'], fmt='npy')
```
ftoschi, line 290: You change from `fmt='npy_pickle'` to `fmt='npy'`, does it work anyway?

wfsim/core/s2.py
```diff
@@ -386,10 +398,82 @@ def luminescence_timings_garfield(xy, n_photons, config, resource, confine_posit
...
+    @staticmethod
+    @njit
+    def draw_excitation_times(inv_cdf_list, hist_indices, nph, diff_nearest_gg, d_gas_gap):
+        
+        """
+        Draws the excitation times from the GARFIELD electroluminescence map
+        
+        :param inv_cdf_list:       List of inverse CDFs for the excitation time histograms
+        :param hist_indices:       The index of the histogram which refers to the gas gap
+        :param diff_nearest_gg:    The difference between the gas gap from the map
```
ftoschi, line 411: Add also the `nph` entry in the comment

ftoschi:
I haven't run it yet (if someone else of the reviewer did it and can confirm it work, I am more than happy), but in general it looks really nice and very detailed, especially in the interpolation between different histograms that could also be an improvement for the old version of the GARFIELD model.

wfsim/load_resource.py
JYangQi00: It works when we test it locally

wfsim/core/s2.py
JYangQi00: Will do, thanks!

wfsim/core/s2.py
JYangQi00: Hi Francesco, so there's sort of two different uses in the electron vs light yield. In the light yield, it either uses the s2_correction_map (if using the old method) or the se_gain_map (using the newer update). This gives the number of photons per electron. For the electron yield, it uses the ratio of g2/SE_gain to get extraction efficiency, so this correction is to get the number of electrons which are extracted. The light yield gets the photons per extracted electrons. I hope this makes sense.

KaraMelih:
For the SE gain map, I think we have a redundant approach.
For every other map, we use the map if it is given, or else we do something else.
For the SE gain map, we have first [a flag
](https://github.com/XENONnT/private_nt_aux_files/blob/5cf6580d4b4607e08aa1d5fd5b3bdbc2a16cf430/sim_files/fax_config_nt_sr0_v0.json#L50)
Then based on that flag we read the SE map, but if the flag is False, or the flag is True but the map doesn't exist it won't work.
Also, if we use this on/off approach [this line](https://github.com/XENONnT/WFSim/blob/41c806969d420e977346c12ebddf33444b1c86b9/wfsim/load_resource.py#L239) should read the attribute only if the se gain from map is True. Otherwise, in principle, there might be no map. 
Also [here](https://github.com/XENONnT/WFSim/blob/41c806969d420e977346c12ebddf33444b1c86b9/wfsim/core/s2.py#L234) it already assumes that SE gain are taken from maps. The test will likely to fail here when using XENON1t as detector and there is no attribute for this parameter.

JoranAngevaare:
Hi @JYangQi00 ,
 -  you are also pointing to a non-existent file https://raw.githubusercontent.com/XENONnT/WFSim/b33656ac59d0366ccf3b20ab0686500cfd403cb6/files/XENONnT_se_xy_map_v1_mlp.json
 - The tests are failing because it's _frankly_ a huge mess which configs are used during those tests. [I can imagine the frustration ](https://github.com/XENONnT/strax_auxiliary_files/pull/30). There are many different pointers to which config is used where
 - On the other hand, I think the tests are broken because the 's2_luminescence_model' is not working for the `simple` model I think this could be fixed by reinstating the right map that comes with that model

JYangQi00:
> * you are also pointing to a non-existent file https://raw.githubusercontent.com/XENONnT/WFSim/b33656ac59d0366ccf3b20ab0686500cfd403cb6/files/XENONnT_se_xy_map_v1_mlp.json

Sorry Joran, I'm a bit confused as to why it's pointing to the non-existent file. It seems that file is located here: https://github.com/XENONnT/private_nt_aux_files/blob/master/strax_files/XENONnT_se_xy_map_v1_mlp.json

terliuk (approved):
I have checked gain/extraction efficiency and it looks reasonable after small changes. So I approve that part. Probably someone can have a brief look again at consistency for gas gap/Garfield luminescence for a few points, so we merge it?

wfsim/core/s2.py
```diff
@@ -354,39 +354,38 @@ def luminescence_timings_simple(xy, n_photons, config, resource):
         return S2._luminescence_timings_simple(len(xy), dG, E0, 
                                                r, dr, rr, alpha, uE,
                                                pressure, n_photons)
-
     @staticmethod
-    def luminescence_timings_garfield(xy, n_photons, config, resource, confine_position=None):
+    @njit
```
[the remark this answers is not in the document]
JYangQi00: I don't see that numba is used on these lines either

ramirezdiego:
@jingqiangye Do we want to keep your last commit? If I understood correctly, the additional population you observed when producing SEs was coming purely from APs, and not from your sampling.

JYangQi00:
> @jingqiangye Do we want to keep your last commit? If I understood correctly, the additional population you observed when producing SEs was coming purely from APs, and not from your sampling.

Hi Diego, I think you meant to tag me. Yeah let's keep it. That tail is still an artifact it seems.

ramirezdiego (approved):

PR #371, opened 2022-05-16
Allow manual setting of gains

JoranAngevaare:
Allow a list of gains as a valid config. See https://github.com/XENONnT/straxen/pull/1039 for an example.

This PR is another solution to the problem described there


JoranAngevaare:
codefactor is complaining about #370

PR #382, opened 2022-06-09
Make release v1.0.0

terliuk:


ramirezdiego (changes requested):
Hey @terliuk, we usually never list the dependabot commits in the release notes/history.

ramirezdiego:
See https://github.com/XENONnT/WFSim/pull/383

Issue #384, opened 2022-06-13
AssertionError - Outside the TPC


PavelKavrigin:
@terliuk @ramirezdiego : The way the interaction positions from **epix** instructions are checked sometimes leads to an assertion error even though the interactions are within the TPC:
[https://github.com/XENONnT/WFSim/blob/69e7095d0385f124291e37d0c108167bb812fa61/wfsim/strax_interface.py#L681](https://github.com/XENONnT/WFSim/blob/69e7095d0385f124291e37d0c108167bb812fa61/wfsim/strax_interface.py#L681) 
[https://github.com/XENONnT/WFSim/blob/69e7095d0385f124291e37d0c108167bb812fa61/wfsim/strax_interface.py#L858](https://github.com/XENONnT/WFSim/blob/69e7095d0385f124291e37d0c108167bb812fa61/wfsim/strax_interface.py#L858)

Here's one the files which led to a crash of WFSim with AssertionError during **mc_chain** run:
[/dali/lgrandi/pkavrigin/2022-06-13_WFSim_XYAssertionError/nT_mc_th228_aa_00188.root](/dali/lgrandi/pkavrigin/2022-06-13_WFSim_XYAssertionError/nT_mc_th228_aa_00188.root)

These are the corresponding X and Y positions from **epix**:
[/dali/lgrandi/pkavrigin/2022-06-13_WFSim_XYAssertionError/nT_mc_th228_aa_00188.X.npy](/dali/lgrandi/pkavrigin/2022-06-13_WFSim_XYAssertionError/nT_mc_th228_aa_00188.X.npy)
[/dali/lgrandi/pkavrigin/2022-06-13_WFSim_XYAssertionError/nT_mc_th228_aa_00188.Y.npy](/dali/lgrandi/pkavrigin/2022-06-13_WFSim_XYAssertionError/nT_mc_th228_aa_00188.Y.npy)

The issue can be reproduced with this notebook:

[/dali/lgrandi/pkavrigin/2022-06-13_WFSim_XYAssertionError/2022-06-13_WFSim_XYAssertionError.ipynb](/dali/lgrandi/pkavrigin/2022-06-13_WFSim_XYAssertionError/2022-06-13_WFSim_XYAssertionError.ipynb)

This is apparently a floating-point precision issue.

PR #386, opened 2022-06-23
Fix R outside due to float32 rounding

terliuk:

Fix for bug where instructions are flagged outside of the TPC due to the float32 rounding. Fixes https://github.com/XENONnT/WFSim/issues/384


ramirezdiego:
Shouldn't you also modify these lines?
https://github.com/XENONnT/WFSim/blob/69e7095d0385f124291e37d0c108167bb812fa61/wfsim/strax_interface.py#L858-L860

PavelKavrigin:
@terliuk - I do not see the assertion error anymore when I use this branch and "undo_1mm_context" branch of **cutax**, with the same example input file which I used for reporting the issue.

PR #387, opened 2022-06-27
Add data-driven map

shenyangshi:
This skeleton change intends to accommodate the data-driven diffusion map, do you have a better idea to nest the simulation/data-driven map neatly and nicely, given the different grid point density?


ramirezdiego:
Linking here the corresponding file upload: https://github.com/XENONnT/private_nt_aux_files/pull/213

terliuk (approved):

PR #412, opened 2023-05-24
Observed interaction position fix

JYangQi00:
Previously, there was an issue in the electron yield function of the S2 simulation, wherein the data-driven S2(x,y) correction maps took the interaction position as input, rather than the observed position (i.e. the position when applying the inverse field-distortion correction). This PR fixes this, and now the observed position is taken as the input to the S2(x,y) correction map.
**WFSim needs you:**

*Pay attention:*


ramirezdiego (approved):
It looks good. I think nothing is missing, but I tag @jarongrigat for him to check & test.

jarongrigat (approved):

JYangQi00:
Thanks Diego and Jaron!

PR #414, opened 2023-07-28
Make runnable notebboks

shenyangshi:
Make existing notebooks runnable including an example of WIMP simulation.


ramirezdiego (approved):
Thanks :)

PR #426, opened 2023-12-20
Update readthedocs configuration

dachengx:

Fix the issue that the docs can not built by RTD, like: https://readthedocs.org/projects/wfsim/builds/22411124/.


ramirezdiego (approved):
Thanks! :)

PR #430, opened 2024-01-15
Move all simulation contexts to WFSim

dachengx:

Together with https://github.com/XENONnT/straxen/pull/1320, to disentangle straxen and wfsim, and make the software environment easier. Because we do not need to maintain compatibility of wfsim inside both wfsim and straxen, e.g. the nestpy(needed by wfsim) currently does not support py3.11(https://github.com/NESTCollaboration/nestpy/issues/104), so we can not update straxen to support py3.11, which is not reasonable.

**WFSim needs you:**

*Pay attention:*

### DIFF
--- a/.github/workflows/Pipi.yml
+++ b/.github/workflows/Pipi.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Setup python
         uses: actions/setup-python@v4
         with:
-          python-version: '3.8'
+          python-version: '3.9'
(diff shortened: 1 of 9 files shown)


ramirezdiego (approved):
Thanks, @dachengx. Let's wait for the pypi release fixes to merge.

WenzDaniel (approved):

Issue #425, opened 2023-12-14
S2 Photon timing downsampled for large time in instruction


FaroutYLq:
When taking a instruction with huge timestamp (also in int64), you will have weird S2 waveforms, where the photons timing are downsampled. More discussion [here](https://xenonnt.slack.com/archives/C0172947DFW/p1702512103588299). It is true to all S2s simulated.

Example for simulation using big time instruction
```python
import cutax
st = cutax.xenonnt_sim_base(output_folder='/scratch/midway3/yuanlq/wfsim')
st.set_config({"fax_file" : "/project/lgrandi/yuanlq/wimp/s2sim/021952-9-se-all-20.csv"}) # feel free to check
se_peaklets = st.get_array('021952_se', 'peaklets')

for p in se_peaklets[:100]:
    if p['dt']==10:
        plt.plot(p['data'])
plt.xlabel('Sample [10ns]')
```
![image](https://github.com/XENONnT/WFSim/assets/47046530/1119e609-82ac-42fd-818b-e3f36590814d)

Example for simulation using small time instruction (what people usually do)
```python
st = cutax.xenonnt_sim_base(output_folder='/scratch/midway3/yuanlq/wfsim')
st.set_config({"fax_file" : "/project/lgrandi/yuanlq/wimp/s2sim/jy_se_f6191.csv"}) # the first 100 rows of /project/lgrandi/jyangqi/se_sim_stuff/se_f6191.csv

se_peaklets = st.get_array('se_f6191', 'peaklets')
for p in se_peaklets[:100]:
    if p['dt']==10:
        plt.plot(p['data'])
plt.xlabel('Sample [10ns]')
```
![image](https://github.com/XENONnT/WFSim/assets/47046530/9d213471-a5ba-4a32-ad23-e6aa8add42d5)

The issue is suspected to happen when int64 were casted into float.

ramirezdiego:
Solution in https://github.com/XENONnT/fuse/pull/171.

PR #434, opened 2024-03-20
Fix s2 photon timing bug

FaroutYLq:
Inspired by solution in fuse [here](https://github.com/XENONnT/fuse/pull/171), trying to solve the [s2 photon timing bug](https://github.com/XENONnT/WFSim/issues/425). The detailed mechanism why it fixed things is unclear but we don't care much for a legacy package. Below is before and after the fix:

image
image

**WFSim needs you:**

*Pay attention:*

### DIFF
--- a/wfsim/core/s2.py
+++ b/wfsim/core/s2.py
@@ -541,7 +541,7 @@ class S2(Pulse):
         # Optical Propagation Delay
         if "optical_propagation" in config['s2_time_model']:
             # optical propagation splitting top and bottom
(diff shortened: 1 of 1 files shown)


ramirezdiego:
Funny:
> FAILED tests/test_wfsim.py::test_sim_1T - ValueError: Found n_hits less than tight_coincidence
============ 1 failed, 20 passed, 439 warnings in 779.44s (0:12:59) ============
Error: Process completed with exit code 1.

FaroutYLq:
OK so it somehow passed after running again. I would recommend just merging it then.

